PR #32, opened 2017-07-22
Added resources file and coding style paragraph (resolves #1)

### LINKED ISSUE #1: Style guide for R lectures and courses
See here for initial issue: https://github.com/UofTCoders/council/issues/175

### Initial comments:
Originally from pull request https://github.com/UofTCoders/council/pull/164

@lwjohnst86 :
For the file naming, it might be best to encourage them to follow a standard that has been developed, rather than create their own. Like Google's R style guide https://google.github.io/styleguide/Rguide.xml

@joelostblom :
I agree that the students should follow an already developed style guideline, and I think it is important that we are consistent across the lectures as well.

I think it is trickier to agree on which style to follow, since R does not have official guidelines. I believe Google's recommendations appear to be sound in large, with the exception of using periods in object/variable and column names. This is confusion for me coming from and object oriented language and I would prefer snake case (a.k.a. underscores: variable_name) in variable names and Pascal / upper camel case in column names (ColumnName). But, and I keep saying this, I'm happy to conform to other standards based on advice from the more experienced R-users among us and discussion with the rest of the group.

However, I do believe underscores would be helpful for students if they are ever to read or write code written in another language, where periods have special meaning. This coding style is not without precedence in the R community, e.g. Hadley Wickham's stat 405 course recommends underscores for variable names, as does this R style guide. The most compelling reason to use underscores in our case might be to stay consistent with the recommendations in the tidyverse guidelines (naturally similar to those from stat 405), since we will largely be using packages from the tidyverse and it would be confusing to mix these conventions with those from another style guide. The tidyverse favors underscores for objects, functions, and parameters, e.g. tbl_df, group_by, geom_point, facet_wrap, etc. The tidyverse style guide is the most exhaustive of the ones I have seen, and often includes justifications of the chosen style, which helps understanding and memorization. There is also a package, lintr, for automatically controlling for adherence to the tidyverse guidelines, and I believe this is the syntax checker that is builtin to Rstudio.

Some additional info: Bioconductor recommends lower camel case. Some Swedish research shows that all three styles are common in existing CRAN packages. This is a helpful SO question, with several good answers.

lwjohnst86:
See discussion at #1.

Any comments on the bit about resubmitting and late penalties?

### DIFF
--- a/resources.Rmd
+++ b/resources.Rmd
@@ -1,0 +1,15 @@
+---
+title: "Resources and Additional Information"
+---
+
+## R Coding Style Guide
+
+*All code submitted for grading* will need to conform to the coding style
+guidelines as outline on [this Style Guide](http://style.tidyverse.org/index.html).
+We will be discussing and highlighting how to automatically (or at least
+consistently) use this style as we go through the lectures and labs. If you do
+not follow these guidelines, you will be requested to make the necessary changes
+and resubmit the assignment. The resubmission *still adheres* to the assignment
+deadline and to the *late submission penalty (5% mark/day)*. So for example, if
+we check that the coding style is not followed one day after the deadline and
+you resubmit one day later, that is a *total of 10% penalty* on that assignment.
--- a/resources.Rmd
+++ b/resources.Rmd
@@ -7,9 +7,8 @@ title: "Resources and Additional Information"
 *All code submitted for grading* will need to conform to the coding style
 guidelines as outline on [this Style Guide](http://style.tidyverse.org/index.html).
 We will be discussing and highlighting how to automatically (or at least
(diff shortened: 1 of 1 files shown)


QuLogic:
Resubmitting sounds like too much work for both of us, especially if we tell them how to create consistent formatting (semi- or completely-)automatically. It depends on the course, but some just have style as part of the mark for every assignment.

lwjohnst86:
Fair enough. What about in earlier assignments we left it pass with a warning and with later assignments it's a requirement?

lwjohnst86:
Fair enough. What about in earlier assignments we left it pass with a warning and with later assignments it's a requirement?

PR #33, opened 2017-07-22
changed table styling (light gray odd rows, border lines)

lwjohnst86:
Table rows now have a very light gray for odd rows. Tables have borders on top and bottom for the header row and the bottom border for the last row.

Added MathJax support for nice (future) equations.

### DIFF
--- a/_output.yml
+++ b/_output.yml
@@ -9,4 +9,4 @@ html_document:
         after_body: _includes/footer.html
     theme: lumen # lumen or simplex or sandstone
     # highlight: pygments # or textmate or tango or default?
-    mathjax: "http://example.com/mathjax/MathJax.js?config=TeX-AMS-MML_HTMLorMML"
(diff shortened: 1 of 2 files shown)


_output.yml
```diff
@@ -5,7 +5,8 @@ html_document:
         collapsed: false
         smooth_scroll: false
     include:
+        in_header: _includes/style.html
         after_body: _includes/footer.html
     theme: lumen # lumen or simplex or sandstone
     # highlight: pygments # or textmate or tango or default?
-    #css: _includes/style.css
+    mathjax: "http://example.com/mathjax/MathJax.js?config=TeX-AMS-MML_HTMLorMML"
```
QuLogic: Using a non-existent domain?

lwjohnst86: huh, weird. I swear I used that link before.. Oh well, fixed it

PR #36, opened 2017-07-23
Commands to generate pdf files to put on Blackboard

lwjohnst86:
I've put together commands to create pdf files for uploading to blackboard.

This is currently incomplete though. I'm not sure how I'll make it more automated to generate these pdfs and (maybe) push to a `pdfs` branch or something. Travis I don't think will be good for this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,5 @@
 .RData
 .Ruserdata
 _site/
+_pdf/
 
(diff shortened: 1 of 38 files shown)


joelostblom:
@lwjohnst86 Is this where you will upload the commands you made for generating the PDFs today also?

lwjohnst86:
This script can convert the syllabus and assignments to PDFs. Let me know if anyone wants to convert their lesson to a PDF, and I can change that. Otherwise, I'll merge this in.

joelostblom:
I don't think there is a requirement for even handing out the lecture content, so it should be good with assignments and syllabus. We can add lecture PDFs later if we think there is a need for it.

joelostblom (approved):

PR #37, opened 2017-07-23
Script to create html slides based on the Rmd files

lwjohnst86:
Started this so we can discuss what we want the slides to look like (if we are even going to use slides).

The way I envision it is that the same Rmd file will be used for the website, pdf for blackboard (see #36), and for the slides. That way there is less overhead for each of us when developing the lessons.


lwjohnst86:
I've updated the slides generating command. So whenever the website gets updated, the slides will get built as well. They'll also be accessible from the website and the GitHub repo (under gh-pages branch).

I'll merge this in. Any suggested stylistic changes to the slides I'll add it to a new PR.

PR #52, opened 2017-08-16
Determine final assessment date

joelostblom:


joelostblom:
Will make one big request instead

PR #68, opened 2017-08-30
Moved tidy data concept up

lwjohnst86:
It doesn't flow very nicely where it was previously.


joelostblom:
Are you thinking of it like an intro to tidy data? Saying what it is and briefly showing how to do it, but mostly staying with already tidied data frames? I am asking because I think too much reshaping can get quite complex quickly.

lwjohnst86:
Yes, definitely only the overall concepts of tidy data, since dplyr emphasizes that concept. (I think maybe even tidyr would be useful to include since they play nicely off of each other)

joelostblom:
Ok I will include it and then we can adjust based on your feedback.

joelostblom:
I have included `gather` and `spread` in assignment 3 and will have them in the lecture as well.

PR #70, opened 2017-08-31
Moved "more on GitHub" up in schedule

lwjohnst86:
Since they will be working on their projects after my session (Oct 17), it would be a good idea to go into more depth with the GitHub/Git. So I moved the GitHub/Git/Scientific Collaboration right after my last session.

@QuLogic what do you think? Is that fine with you?

Others?


joelostblom (approved):
It makes sense to move the GitHub collaboration lecture. I might want to add a project work session after the GitHub session and move Lina's class until after fall break. But I am not sure if this is a good idea, since your assignment already cover most of the things they should be doing at this point so it might not be necessary and its better to save the project work sessions until later

joelostblom:
Woops, sorry @QuLogic, please provide feedback whether this is ok for you and we can change accordingly.

joelostblom:
Ah ok good, I thought I merged it already. all good then

PR #78, opened 2017-09-06
Fix some small typos.

QuLogic:
I tried to fix the rubric, but I don't know why it renders so poorly on the website vs locally, so I left it as is.


lwjohnst86:
I know why. The column width is determined by the width of the `|---|` portions. So, to fix that, just do something like (I had been planning on doing that tomorrow, but since you've started this...):

```
|-----------|-------------------------------------------------------|-------------------------------------------------------|-------------------------------------------------------|
```

QuLogic:
Well, that looked the same before and after, locally. Hopefully it works here.

joelostblom:
Thanks @QuLogic, I wil merge it so that we can see how it looks on the website.

PR #80, opened 2017-09-06
Try to fix table formatting again

joelostblom:
Merging right away to render on the website


joelostblom:
@QuLogic @lwjohnst86  The final project table looks good in fullscreen now. In halfscreen, the rightmost column is a bit compressed, but we would need to specify some type of evenly divided reflow between column to fix that, instead of just removing width from the last column when the window gets narrower. I don't think this is important at all, especially as reflow looks good on mobile devices, it's only narrow windows on a computer that look a little bit wonky.

PR #91, opened 2017-09-09
Add lecture 2

joelostblom:


lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+We will be working with the programming language R through an interface called
+RStudio. We will mainly be using the R Notebook, which allows us to mix R code
+with writing notes in the Markdown text format. So let's open an R Notebook in
+RStudio... You will be prompted to install a few packages, click Yes/OK. I have
+asked the IT services to make sure these packages are always installed on the
+computers so that we don't have to install them every time, hopefully they will
+do so soon. If you prefer, you can also bring your own laptop to class, which
+has the advantage that you can save things permanently and that you have
+everything setup, this is up to you <!--ok to say this?-->. Let's go over the
```
joelostblom: Is it ok to offer them to bring their own computers? Any unequality that could arise from that? Should not since they all have access to computers right?

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+of a single type, e.g. all numbers. Above, we saw 2 of the 6 main **atomic
+vector** types  that R uses: `"character"` and `"numeric"` (or `"double"`).
+These are the basic building blocks that all R objects are built from. The other
+4 **atomic vector** types are:
+
+* `"logical"` for `TRUE` and `FALSE` (the boolean data type)
+* `"integer"` for integer numbers (e.g., `2L`, the `L` indicates to R that it's an integer)
+* `"complex"` to represent complex numbers with real and imaginary parts (e.g.,
+  `1 + 4i`) and that's all we're going to say about them
+* `"raw"` for bitstreams that we won't discuss further
```
joelostblom, line 470: Not sure if I should even mention this or just introduce logicals, since they will use them for conditional subsetting. Same for the challenge just below.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+```{r}
+animals <- c("mouse", "rat", "dog", "cat")
+animals[animals == "cat" | animals == "rat"] # returns both rat and cat
+animals %in% c("rat", "cat", "dog", "duck", "goat")
+animals[animals %in% c("rat", "cat", "dog", "duck", "goat")]
+```
+
+> ### Challenge
+>
+> * Can you figure out why `"four" > "five"` returns `TRUE`?
```
joelostblom: Maybe add a question on conditional subsetting

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+```{r}
+## Answers
+## * When using ">" or "<" on strings, R compares their alphabetical order. Here
+##   "four" comes after "five", and therefore is "greater than" it.
+```
+-->
+
+
+## Missing data
+
```
joelostblom, line 596: Missing data is indeed important, but how important are these techniques when working with `dplyr`?

joelostblom:
@mbonsma @linamnt Adding you guys as reviewers for the lecture tomorrow, since I don't want to overload @lwjohnst86 to be the only one to always review my R lectures. This lecture is also pretty introductory, so it is almost more general coding principles than R specific syntax.

joelostblom:
Should there be more time for them to code themselves, aka more challenges?

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+When assigning a value to an object, R does not print anything. You can force R
+to print the value by using parentheses or by typing the object name:
+
+```{r}
+weight_kg <- 55    # doesn't print anything
+(weight_kg <- 55)  # but putting parenthesis around the call prints the value of `weight_kg`
+weight_kg          # and so does typing the name of the object
+```
+
+The variable `weight_kg` is not stored in the computer's memory where R can
```
linamnt: do you mean "`weight_kg` is stored..."?

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+```{r, eval=FALSE}
+round(3.14159, digits = 2)
+```
+
+So, above we pass the *argument* `2`, to the *parameter* `digits`. Knowing this
+nomenclature is not essential for doing your own data analysis, but it will be
+very helpful when you are reading through help documents online and in RStudio.
+
+We can leave actually, out the word `digits` since we know it comes as the
```
linamnt: "We can leave out the word..." take out "actually,"

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+  `1 + 4i`) and that's all we're going to say about them
+* `"raw"` for bitstreams that we won't discuss further
+
+Vectors are one of the many **data structures** that R uses. Other important
+ones are lists (`list`), matrices (`matrix`), data frames (`data.frame`),
+factors (`factor`) and arrays (`array`). In this class, we will focus on data
+frames, which is most commonly used one for data analyses.
+
+> ### Challenge
+> * We’ve seen that atomic vectors can be of type character, >   numeric (or
```
linamnt: you have a ' > ' where you shouldn't after "character, ..."

lec02-basic-r.Rmd
linamnt: i think it's fine for now until you eventually introduce logicals...?

lwjohnst86:
Yes, they should totally have time to do a couple exercises to code on their own.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+*Note: This lecture content was originally created by voluntary contributions to
+[data carpentry](http://datacarpentry.org) and has been modified to
+align with the aims of EEB430. Data Carpentry is an organization focused on data
+literacy, with the objective of teaching skills to researchers to enable them to
+retrieve, view, manipulate, analyze and store their and other’s data in an open
+and reproducible way in order to extract knowledge from data. For EEB430, we are
+making all our content available under the same license, The Creative Commons, 
+license, so if someone in the future are teaching this course content available under
+the same lice*
```
lwjohnst86: lice? ew :P `s/lice/license/`

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
 ---
-title: "Basic R (part 1): Rstudio, Rmarkdown, assignment, vectors"
+title: Introduction to R
 ---
 
+**Copyright (c) Data Carpentry**
+
+*Note: This lecture content was originally created by voluntary contributions to
+[data carpentry](http://datacarpentry.org) and has been modified to
```
lwjohnst86: Data Carpentry

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+align with the aims of EEB430. Data Carpentry is an organization focused on data
+literacy, with the objective of teaching skills to researchers to enable them to
+retrieve, view, manipulate, analyze and store their and other’s data in an open
```
lwjohnst86: Oxford comma: `analyze, and store`

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+and reproducible way in order to extract knowledge from data. For EEB430, we are
+making all our content available under the same license, The Creative Commons,
```
lwjohnst86: Link to Creative Commons? https://creativecommons.org/

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+**Copyright (c) Data Carpentry**
+
+*Note: This lecture content was originally created by voluntary contributions to
+[data carpentry](http://datacarpentry.org) and has been modified to
+align with the aims of EEB430. Data Carpentry is an organization focused on data
+literacy, with the objective of teaching skills to researchers to enable them to
+retrieve, view, manipulate, analyze and store their and other’s data in an open
+and reproducible way in order to extract knowledge from data. For EEB430, we are
+making all our content available under the same license, The Creative Commons, 
+license, so if someone in the future are teaching this course content available under
```
lwjohnst86: `s/future are/future will be/`

lwjohnst86: > ... so if that anyone in the future can re-use or modify our course content, without infringing on copyright licensing issues.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+the same lice*
+
+I am emphazing this because I think it is one of the core features of working
```
lwjohnst86: I would strongly recommend you change from `I` to `we` (since we think/believe the same thing). Or better yet, try to minimize first/plural person as much as you can while still maintaining an [active voice](https://webapps.towson.edu/ows/activepass.htm)

lwjohnst86: `emphasizing this` what is `this` in this context? It's not entirely clear.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+with an open language like R, many smart people readily share their material and
```
lwjohnst86: > with an open language like R. Many smart people willingly and actively share their material publicly and in the open, so that others can modify and build off of the material themselves.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+discoveries not only for free, but in a way that allows others to modify and
+built on the material themselves. It really is "standing on the shoulders of
```
lwjohnst86: The 'stand on shoulders' and 'stayed for the community' are two separate topics and don't flow together when in the same sentence. Maybe:

> By being open, we can "stand on the shoulders of giants" and to continue to contribute for others to then stand on our shoulders. Not only does this help get work done, but it also adds to a feeling of community. In fact, there is a common saying in the open source world: ...

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+giants", and there is a common saying in the open source world:
+
+> I came for the language and stayed for the community.
+
+Which highlights the spirit, generousity, and the fun involved in many of these project.
+
+------------
+
+
+## Recap from intro lecture
```
lwjohnst86: This section is useful to say and discuss in class, but not to include in the lesson material.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+I am emphazing this because I think it is one of the core features of working
+with an open language like R, many smart people readily share their material and
+discoveries not only for free, but in a way that allows others to modify and
+built on the material themselves. It really is "standing on the shoulders of
+giants", and there is a common saying in the open source world:
+
+> I came for the language and stayed for the community.
+
+Which highlights the spirit, generousity, and the fun involved in many of these project.
```
lwjohnst86: > This saying highlights the spirit, generosity, and the fun involved in being a part of and participating in these open source projects.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+* Define the following terms as they relate to R: object, assign, call,
+  function, arguments, options.
+* Create objects and assign values to them.
+* Use comments to inform script.
+* Do simple arithmetic operations in R using values and objects.
+* Call functions and use arguments to change their default options.
+* Inspect the content of vectors and manipulate their content.
+* Subset and extract values from vectors.
+* Correctly define and handle missing values in vectors.
+
```
lwjohnst86: Please include a lesson outline with approximate timing.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+## Setting up the notebook
```
lwjohnst86: `R Notebook`

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+Let's remove the template R Studio gives us, and add a title of our own.
```
lwjohnst86: RStudio

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+```yaml
+---
+title: Introduction to R
+---
+```
+
+This introductory block is call the `yaml`-block and is where we specify things
```
lwjohnst86: > This header block is called the YAML header. This is where we specify whether we want to convert this file to a HTML or PDF file. This will be discussed in more detail in another class.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+like whether we want html or PDF output, but more on that another class. For
```
lwjohnst86: > For now, let's type out a note:

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+now, we just care about including the lecture here. Under it let's type a note:
+
+> This lecture covers fundamental R usage, such as assigning values to a
+> variable, using functions, commenting code, and more.
+
+Under this sentence, we will insert our first code chunk. You insert a code
+chunk by either clicking this button or pressing <kbd>Ctrl</kbd> +
```
lwjohnst86:
```
clicking 'Insert' button or ....
```

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+<kbd>Alt</kbd> + <kbd>i</kbd>. You could also type out the surrounding
+delimiters, but this would take longer. To run a code chunk, you press this
```
lwjohnst86: > surrounding backticks (```), but ...

lwjohnst86: > you press the green arrow ...

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+the assignment operator `<-`, and the value we want to assign:
+
+```{r}
+x <- 3
+x + 5
+```
+
+`<-` is the assignment operator. It assigns values on the right to objects on
+the left. So, after executing `x <- 3`, the value of `x` is `3`. The arrow can
+be read as 3 **goes into** `x`.  For historical reasons, you can also use `=`
```
lwjohnst86: I won't bother mentioning using the `=`. I doubt any other student has used another programming language and would understand this. I would also suggest not bothering to mention it since (at this point) the less they need to worry about, the more they will learn what we want them to learn.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+for assignments, but not in every context. Because of
+the
+[slight](http://blog.revolutionanalytics.com/2008/12/use-equals-or-arrow-for-assignment.html) [differences](https://web.archive.org/web/20130610005305/https://stat.ethz.ch/pipermail/r-help/2009-March/191462.html) in
+syntax, it is good practice to always use `<-` for assignments.
```
lwjohnst86: You can remove lines 103-106. If it comes up, we mention it. You could move this to a footnote though...

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+In RStudio, typing <kbd>Alt</kbd> + <kbd>-</kbd> (push <kbd>Alt</kbd> at the
+same time as the <kbd>-</kbd> key) will write ` <- ` in a single keystroke.
+
+What are known as `objects` in `R` are called `variables` in many other
```
lwjohnst86: No need to mention differences between languages. These are beginners. This is an interesting fact for someone who is a bit intermediate. Less is better in learning.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+programming languages.  Depending on the context, `object` and `variable` can
+have drastically different meanings.  However, in this lesson, the two words are
+used synonymously and in this introductory section, the title "Creating object
+in R" is the same as "Assigning a value to a variable" (for more information
```
lwjohnst86: Why don't you just say `Assigning values to variables` for the section title?

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+see: https://cran.r-project.org/doc/manuals/r-release/R-lang.html#Objects).
```
lwjohnst86: Put this as a footnote instead.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+You can name an object in R, almost anything you want:
```
lwjohnst86: > ... to almost anything you want:

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+```{r}
+joel <- 3
+joel + 5
+```
+
+> ### Challenge
```
lwjohnst86: Use bold here, not a header, since it will mess with the table of contents.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+> So far, we have created two variables, `joel` and `x`. What is the sum of these variables?
+
+> ```{r}
+> joel + x
+> ```
+
+Objects can be given any name such as `x`, `current_temperature`, or
+`subject_id`. You want your object names to be explicit and not too long. They
```
lwjohnst86, line 137: You could link to the resources page here, on coding style (https://uoftcoders.github.io/rcourse/resources.html)

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+cannot start with a number (`2x` is not valid, but `x2` is). R is case sensitive
+(e.g., `joel` is different from `Joel`). There are some names that cannot be
+used because they are the names of fundamental functions in R (e.g., `if`,
+`else`, `for`, see
+[here](https://stat.ethz.ch/R-manual/R-devel/library/base/html/Reserved.html)
+for a complete list). In general, even if it's allowed, it's best to not use
+other function names (e.g., `c`, `T`, `mean`, `data`, `df`, `weights`). If in
+doubt, check the help or use tab completion to see if the name is already in
+use. It's also best to avoid dots (`.`) within a variable name as in
+`my.dataset`. There are many functions in R with dots in their names for
```
lwjohnst86:
```
Historically, there are many functions in R with dots in their name, but since dots have a [special meaning](http://adv-r.had.co.nz/S3.html) in R, it's better to not use them and instead use underscores (`_`).
```

Just removed mention of "methods" and "other programming languages". No need to get into that for beginners.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+historical reasons, but because dots have a special meaning in R (for methods)
+and other programming languages, it's safest to avoid them and use underscores
+(`_`) instead. It is also recommended to use nouns for variable names, and verbs
+for function names. It's important to be consistent in the styling of your code
+(where you put spaces, how you name variables, etc.). Using a consistent coding
+style makes your code clearer to read for your future self and your
+collaborators. In R, three popular style guides are
```
lwjohnst86: Just refer to our resource page for this and don't go into too much detail. They don't need to know that there are other style guides. They need to use ours (or rather tidyverse :P )

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+[Google's](https://google.github.io/styleguide/Rguide.xml), [Jean
+Fan's](http://jef.works/R-style-guide/) and the
+[tidyverse's](http://style.tidyverse.org/). The tidyverse's is very
+comprehensive and may seem overwhelming at first. You can install the
+[**`lintr`**](https://github.com/jimhester/lintr) to automatically check for
+issues in the styling of your code.
```
lwjohnst86: RStudio already comes with a linter. The key binding is `Ctrl-Shift-A` for highlighted code.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+weight_lb <- 2.2 * weight_kg
+```
+
+and then change `weight_kg` to 100.
+
+```{r}
+weight_kg <- 100
+```
+
+> ### Challenge
```
lwjohnst86: See note above. Use **bold** instead of header

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+> ```{r}
+> weight_lb
+> ```
+
+
+### Comments
+
+A comment is piece of text in your code that will not be executed with the rest
+of the code, but its purpose is to be a note to the person reading the code.
+Traditionally, comments has been the only way to make notes when you are coding,
```
lwjohnst86: > comments have been

lwjohnst86: > to make notes along side your code,

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+but since we are using the R Notebook format, we can also write notes in
+markdown. Therefore, comments are not as important for us, but they are still
```
lwjohnst86: s/markdown/Markdown

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+<kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>C</kbd>. If you only want to comment
+out one line, you can put the cursor at any location of that line (i.e. no need 
+to select the whole line), then press <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + 
+<kbd>C</kbd>.
+
+```{r}
+weight_lb <- 2.2 * weight_kg # Actually, 1 kg = 2.204623 lbs
+```
+
+> ### Challenge
```
lwjohnst86: See note above about bolding

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+```
+
+Here, the value of `a` is given to the `sqrt()` function, the `sqrt()` function
+calculates the square root, and returns the value which is then assigned to
+variable `b`. This function is very simple, because it takes just one argument.
+
+The return 'value' of a function need not be numerical (like that of `sqrt()`),
+and it also does not need to be a single item: it can be a set of things, or
+even a dataset. We'll see that when we read data files into R, as I showed you
+briefly in during the first class.
```
lwjohnst86: > it can be a set of things, or even a dataset, as we'll see later on.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+To access help about `sqrt`, we are first going to learn about tab-completion.
+Type `s` and press <kbd>Tab</kbd>.
+
+```{r, eval=FALSE}
+s<tab>q
+```
+
+You can see that R gives you suggestions of what functions and variables are
+available that start with the letter `s`, and thanks to `RStudio` they are
```
lwjohnst86: > ... thanks to RStudio

No need to use backticks

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+formatted in this nice list. There are *many* suggestions here, so let's be a
+bit more specific and append a `q`, to find what we want. If we press enter or
+tab again, R will insert the selected option.
+
+You can see that R inserts a pair of parenthesis together with the name of the
+functions. This is how the function syntax looks for R and many other
+programming languages, and it means that within these parentheses, we will
+specify all the arguments, the ingredients, that we want to pass to this
```
lwjohnst86: > all the arguments (the ingredients) that ...

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+function.
+
+If we press tab again, R will helpfully display all the available parameters for
+this function that we can pass an argument to. The word *parameter* is used to
+describe the name that the argument can be passed to, more on that later.
```
lwjohnst86: > can be passed to. More on that later.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+```{r, eval=FALSE}
+sqrt(<tab>
+```
+
+There are many things in this list, but only one of them is marked in purple.
+Purple here means that it is parameter we can use for the function and yellow
+means that it is a variable that we defined earlier. There are a few other
+symbols as well, all of which can be viewed at the end of [this post about
+RStudio code
+completion](https://support.rstudio.com/hc/en-us/articles/205273297-Code-Completion).
```
lwjohnst86: This might be better as a footnote

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+To read the full help about `sqrt`, we can use the question mark, or type it
+directly into the help document browser.
+
+```{r}
```
lwjohnst86: `eval=FALSE`

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+?sqrt
+```
+
+As you can see, `sqrt()` takes only one argument, `x`, which needs to be a
+*numerical vector*. Don't worry too much about that is says *vector* here, I
+will talk more about that later. But briefly, a vector is a set of one or
+several numbers, and in R, every value is a vector even if it only is a single
+number. So, you don't have to do anything special in R to create a vector, but
+more on that later.
```
lwjohnst86: >... numerical vector. Briefly, a vector is one or more numbers. Every number is a vector, so you don't have to do anything special to create a vector. More on vectors later.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+```{r}
+round(3.14159)
+```
+
+Here, we've called `round()` with just one argument, `3.14159`, and it has
+returned the value `3`. That's because the default is to round to the nearest
+whole number, or integer. If we want more digits we can pass an argument to the
+`digits` parameter, to specify how many decimals we want to round to.
+
+```{r, eval=FALSE}
```
lwjohnst86: why eval = false?

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+* `"complex"` to represent complex numbers with real and imaginary parts (e.g.,
+  `1 + 4i`) and that's all we're going to say about them
+* `"raw"` for bitstreams that we won't discuss further
+
+Vectors are one of the many **data structures** that R uses. Other important
+ones are lists (`list`), matrices (`matrix`), data frames (`data.frame`),
+factors (`factor`) and arrays (`array`). In this class, we will focus on data
+frames, which is most commonly used one for data analyses.
+
+> ### Challenge
```
lwjohnst86: Bold instead of header.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+> * We’ve seen that atomic vectors can be of type character, >   numeric (or
+double), integer, and logical. But what happens if we try to mix these types in 
+>   a single vector? Find out by using `class` to test these examples?
```
lwjohnst86: no `?` at end needed.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+a search vector are found:
+
+```{r}
+animals <- c("mouse", "rat", "dog", "cat")
+animals[animals == "cat" | animals == "rat"] # returns both rat and cat
+animals %in% c("rat", "cat", "dog", "duck", "goat")
+animals[animals %in% c("rat", "cat", "dog", "duck", "goat")]
+```
+
+> ### Challenge
```
lwjohnst86: bold instead

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+## Extract those elements which are not missing values.
+heights[!is.na(heights)]
+
+## Returns the object with incomplete cases removed. The returned object is an atomic vector of type `"numeric"` (or `"double"`).
+na.omit(heights)
+
+## Extract those elements which are complete cases. The returned object is an atomic vector of type `"numeric"` (or `"double"`).
+heights[complete.cases(heights)]
+```
+Recall that you can use the `typeof()` function to find the type of your atomic vector.
```
lwjohnst86: typeof wasn't introduced. You mean class right?

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+> ### Challenge
```
lwjohnst86: bold

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+together:
+
+```{r}
+add_two_numbers <- function(num1, num2) {
+    return(num1 + num2)
+}
+add_two_numbers(4, 5)
+```
+
+As you can see, running this function on two numbers, returns their sum. We
```
lwjohnst86: > on two numbers returns their sum

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+```{r}
+add_two_numbers <- function(num1, num2) {
+    my_sum <- num1 + num2
+    return(my_sum)
+}
+add_two_numbers(4, 5)
+```
+
+> ### Challenge
```
lwjohnst86: bold

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+> ```
+
+
+## Loops and vectorization
+
+Loop, or for-loops, are essential to programming in general. However, in R, you
+should avoid them as often as possible because there are more efficient ways of
+doing things that you should use instead. It is still important that you
+understand the concept of loops and you might also use them in some of your own
+functions if there is no vectorized way of going about what you want to to.
```
lwjohnst86: `what you want to do`

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+
+You can think of a for-loop as: "for each number contained in a list/vector,
+perform this operations" and the syntax basically says the same thing:
```
lwjohnst86: > operation

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+my_sum(v)
+```
+
+Although this gives us the same output as the built-in function `sum`, the
+built-in function has many more optimizations so it is much faster than our
+function. In R, it is always faster to try to find a way of doing things without
+writing a loop yourself. When you are reading about R, you might see suggestions
+that you should try to *vectorize* your code to make it faster. What people are
+referring to, is that you should not write for loops in R and instead use the
+ready made function that are much more efficient in working with vectors and
```
lwjohnst86: > ready made functions

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,754 @@
...
+essentially essentially performs operations on entire vector at once instead of
```
lwjohnst86: essentially twice here

lwjohnst86 (changes requested):
Good overall. I've got some feedback within. Mostly it is typographical/grammar/formatting for the website. The content is good!

lec02-basic-r.Rmd
joelostblom: Wow, what a mess this intro was... thanks for cleaning it up.

lec02-basic-r.Rmd
QuLogic: ready-made

lec02-basic-r.Rmd
joelostblom: I'll move this all to the footnote. Some of them said they have programming experience and `=` is what they use in math for assignment so makes sense to include it in the footnote i think.

lec02-basic-r.Rmd
joelostblom: Agree, was going to leave it in without saying it. Footnote is better.

lec02-basic-r.Rmd
joelostblom: I thought `object` was more correct than `variable` in R. I'm happy to change to `variable`. But I think it is worth mentioning `objects` since documentation they read elsewhere might use that term.

joelostblom (commented):
Thanks guy! Will fix the remaining issues tomorrow morning.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+Commons](https://creativecommons.org/), so that anyone in the future
+can re-use or modify our course content, without infringing on copyright
+licensing issues.*
+
+The above paragraph is made explicit since it is one of the core features of
+working with an open language like R. Many smart people willingly and actively
+share their material publicly, so that others can modify and build off of the
+material themselves.
+
+By being open, we can "stand on the shoulders of giants" and to continue to
```
QuLogic: and continue

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+contribute for others to then stand on our shoulders. Not only does this help
+get work done, but it also adds to a feeling of community. In fact, there is a
+common saying in the open source world:
+
+with an open language like R, many smart people readily share their material and
+discoveries not only for free, but in a way that allows others to modify and
+built on the material themselves. It really is "standing on the shoulders of
+giants", and there is a common saying in the open source world:
```
QuLogic: This paragraph seems to be duplicated.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+the console:
+
+```{r}
+3 + 5
+12 / 7
+```
+
+However, to do useful and interesting things, we need to assign _values_ to
+_objects_. Objects are also referred to as _variables_, and although there
+technically are minor differences between them[^objects_variables], we will use them interchangeably
```
QuLogic: s/them/the terms/

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+
+```{r}
+x <- 3
+x + 5
+```
+
+`<-` is the assignment operator. It assigns values on the right to objects on
+the left. So, after executing `x <- 3`, the value of `x` is `3`. The arrow can
+be read as 3 **goes into** `x`.[^equal_vs_arrow] In RStudio, typing
+<kbd>Alt</kbd> + <kbd>-</kbd> (push <kbd>Alt</kbd> at the same time as the
```
QuLogic: The "at the same time" explanation should have gone above where the keyboard shortcuts were first shown.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+
+> ```{r}
+> joel + x
+> ```
+
+Objects can be given any name such as `x`, `current_temperature`, or
+`subject_id`. You want your object names to be explicit and not too long. They
+cannot start with a number (`2x` is not valid, but `x2` is). R is case sensitive
+(e.g., `joel` is different from `Joel`). There are some names that cannot be
+used because they are the names of fundamental functions in R (e.g., `if`,
```
QuLogic: keywords?

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+Arguments can be anything, not only numbers or filenames, but also other
+objects. Exactly what each argument means differs per function, and must be
+looked up in the documentation (see below). Some functions take arguments which
+may either be specified by the user, or, if left out, take on a *default* value:
+these are called *options*. Options are typically used to alter the way the
+function operates, such as whether it ignores 'bad values', or what symbol to
+use in a plot. However, if you want something specific, you can specify a value
+of your choice which will be used instead of the default.
+
+To access help about `sqrt`, we are first going to learn about tab-completion.
```
QuLogic, line 287: Sort of already did this.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+factors (`factor`) and arrays (`array`). In this class, we will focus on data
+frames, which is most commonly used one for data analyses.
+
+> **Challenge**
+>
+> * We’ve seen that atomic vectors can be of type character, numeric (or
+double), integer, and logical. But what happens if we try to mix these types in 
+>   a single vector? Find out by using `class` to test these examples.
+
+> ```r
```
QuLogic: Missing the braces around `r`.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+heights[complete.cases(heights)]
+```
+Recall that you can use the `class()` function to find the type of your atomic vector.
+
+> **Challenge**
+>
+> 1. Using this vector of length measurements, create a new vector with the NAs
+> removed.
+
+> ```r
```
QuLogic: Missing the braces around `r`.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+
+```{r}
+more_animals <- animals[c(1, 2, 3, 2, 1, 4)]
+more_animals
+```
+
+R indices start at 1. Programming languages like Fortran, MATLAB, Julia, and R
+start counting at 1, because that's what human beings typically do. Languages in
+the C family (including C++, Java, Perl, and Python) count from 0 because that's
+simpler for computers to do.
```
QuLogic: I don't think simplicity for computers is the reason.
https://python-history.blogspot.ca/2013/10/why-python-uses-0-based-indexing.html
https://www.cs.utexas.edu/~EWD/transcriptions/EWD08xx/EWD831.html

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+
+```{r}
+weight_g > 50    # will return logicals with TRUE for the indices that meet the condition
+## so we can use this to select only the values above 50
+weight_g[weight_g > 50]
+```
+
+You can combine multiple tests using `&` (both conditions are true, AND) or `|`
+(at least one of the conditions is true, OR), similar to how you can do when
+using a search engine online:
```
QuLogic: With Google becoming "smart", I'm not sure if this is something you really do with search engines any more.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+> ```
+>
+> 2. Use the function `median()` to calculate the median of the `lengths` vector.
+
+
+## Writing functions
+
+In this class, you will be working a lot with functions, especially those that
+someone else has already written. When you type `sum`, `c()`, or `mean()`, you
+are using a function that has be made previously and built into R. To remove
```
QuLogic: has been

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+some of the magic around these functions, we wil go through how to make a basic
```
QuLogic: we will

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+```
+
+Instead of printing out every number to the console, we could also add numbers
+cumulatively, to calculate the sum of all the numbers in the vector:
+
+```{r}
+# To increment `w` each time, we must first create the variable,
+# which we do by setting `w <- 0`, referred to as initializing.
+# This also ensures that `w` is zero at the start of the loop and
+# don't retain the value from last time we ran this code.
```
QuLogic: doesn't

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+```
+
+Although this gives us the same output as the built-in function `sum`, the
+built-in function has many more optimizations so it is much faster than our
+function. In R, it is always faster to try to find a way of doing things without
+writing a loop yourself. When you are reading about R, you might see suggestions
+that you should try to *vectorize* your code to make it faster. What people are
+referring to, is that you should not write for loops in R and instead use the
+ready made functions that are much more efficient in working with vectors and
+essentially performs operations on entire vector at once instead of
```
QuLogic: on the

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,765 @@
...
+one number at a time. If anyone is interested in more details about how this
+work, please ask me after class, but conceptually loops operate on one element
```
QuLogic: works

QuLogic (commented):
A bit of the frontend could be shortened perhaps, depending on how confident students feel with R right now.

lec02-basic-r.Rmd
joelostblom: I will remove this for the final version

lec02-basic-r.Rmd
joelostblom: Will include in final version, timing session today to practice

lec02-basic-r.Rmd
joelostblom: Seems like it was when the initial implementation was done https://en.wikipedia.org/wiki/Zero-based_numbering#Origin. Will add "historically".

joelostblom: Also added that it can allow for more elegant code.

lec02-basic-r.Rmd
mbonsma: But maybe it's still good to do it explicitly? It's an important thing, and if someone misses it it would be bad.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,759 @@
...
+
+There are many functions that allow you to inspect the content of a
+vector. `length()` tells you how many elements are in a particular vector:
+
+```{r}
+length(weight_g)
+length(animals)
+```
+
+An important feature of a vector, is that all of the elements are the same type
```
mbonsma: remove comma here

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,759 @@
...
+
+```{r, echo=FALSE, eval=FALSE}
+num_char <- c(1, 2, 3, "a")
+
+num_logical <- c(1, 2, 3, TRUE)
+
+char_logical <- c("a", "b", "c", TRUE)
+
+tricky <- c(1, 2, 3, "4")
+```
```
mbonsma: What's this R section for?

mbonsma (approved):
Looks great! I'm looking forward to learning all this myself :)

lec02-basic-r.Rmd
lwjohnst86: Not super important but still good to know. Under the hood, most stats functions deal with missing data by default.

lec02-basic-r.Rmd
```diff
@@ -1,4 +1,757 @@
...
+ready-made functions that are much more efficient in working with vectors and
+essentially performs operations on entire vector at once instead of
+one number at a time. If anyone is interested in more details about how this
+works, please ask me after class, but conceptually loops operate on one element
+at a time while vectorized code operate on all elements of a vector at once.
+
+*Go to the output options and set it to `html_document`, and then knit*
+
+**Remember to save your notebook file (.Rmd) on a USB stick or email it to
+yourself**
```
lwjohnst86: These two lines can be deleted. They aren't necessary for the online lesson material, since this is a class/situation dependent issue.

lwjohnst86:
Only need to deal with Maddy's two comments and my one comment, then we can merge it in!

PR #92, opened 2017-09-11
Add lecture 3

joelostblom:
@QuLogic @lcoome I am involving more people in the reviews, so that @lwjohnst86 doesn't get overloaded =)

PR #95, opened 2017-09-13
Fixed slides script to only build specific lectures

lwjohnst86:
This should the CI error in #92


joelostblom (commented):
Don't quite understand how this works. Will you change the `rmd_files` vector manually when you want to build slides for your lectures?

lwjohnst86:
Yes. Since not everyone wants to create slides, it doesn't make sense to generate slides for every since lecture. This would just increase CI build time.

PR #97, opened 2017-09-14
Add assignment 2

joelostblom:


joelostblom:
The content here is accidentally from the `.md` file instead of the `.Rmd` file.

PR #102, opened 2017-09-14
New assignment 2 and assignment 1 update

QuLogic:


QuLogic:
OK'd by @lwjohnst86 here in class.

PR #109, opened 2017-09-19
Add lecture 5

joelostblom:
Will add the second half as a separate pull request when it is ready (probably late tonight)

### DIFF
--- a/lec05-dplyr.Rmd
+++ b/lec05-dplyr.Rmd
@@ -162,7 +162,7 @@ surveys_gw_wide %>%
 >    columns for `spread`.
 
 <!---
-  ```{r, echo=FALSE, purl=FALSE}
+  ```{r, echo=FALSE}
 ## Answer 1
 rich_time <- surveys %>%
   group_by(plot_id, year) %>%
@@ -436,8 +436,8 @@ for the exact colors and shapes, axis styles, or to include the small dots
 around the main lines).
 
 ```{r, eval=FALSE, echo=FALSE}
-plant_biomass %>% 
-    gather('species', 'biomass', betula_nana:vaccinium_vitis_idaea) %>% 
+plant_biomass %>%
+    gather(species, biomass, betula_nana:vaccinium_vitis_idaea) %>% 
     group_by(habitat, treatment, year) %>% 
     summarize(mean_biomass = mean(biomass)) %>% 
     ggplot(aes(x = year, y = mean_biomass, color = treatment)) +
@@ -461,7 +461,7 @@ Hints:
 
 ```{r, eval=FALSE, echo=FALSE, fig.height=9, fig.width=7}
 plant_biomass %>% 
-    gather('species', 'biomass', betula_nana:vaccinium_vitis_idaea) %>% 
+    gather(species, biomass, betula_nana:vaccinium_vitis_idaea) %>% 
     group_by(habitat, species, treatment, year) %>% 
     summarize(mean_biomass = mean(biomass)) %>% 
     ggplot(aes(x = year, y = mean_biomass, color = treatment)) +
@@ -488,7 +488,7 @@ and search online to find out how to:
 color_names <- c('red', 'blue')
 
 plant_biomass %>% 
-    gather('species', 'biomass', betula_nana:vaccinium_vitis_idaea) %>% 
+    gather(species, biomass, betula_nana:vaccinium_vitis_idaea) %>% 
     group_by(habitat, treatment, year) %>% 
     summarize(mean_biomass = mean(biomass)) %>% 
     ggplot(aes(x = year, y = mean_biomass, color = treatment, shape = treatment)) +
--- a/lec05-dplyr.Rmd
+++ b/lec05-dplyr.Rmd
@@ -3,17 +3,17 @@ title: "Tidying and exporting data, reproducing figures"
 ---
 
 > ### Lecture objectives
-> 
+>
 > - Learn about tidy data.
 > - Transform data from the long to wide format.
 > - Reproduce existing figures from raw data.
 > - Understand which raw data is underlying a figure.
 > - Understand which types of figures are suitable to create from raw data.
 > - Explore scientific questions using dplyr and ggplot.
-> 
-> 
+>
+>
 > ### Lecture outline
-> 
+>
 > - Reshaping with gather and spread (25 min)
 > - Exporting data (15 min)
 > - Reproducing figures (50 min)
@@ -229,7 +229,7 @@ if (!dir.exists('data')) {
 }
 ```
 
-We are going to prepare a cleaned up version of the data without NAs. Let's 
+We are going to prepare a cleaned up version of the data without NAs. Let's
 start by removing observations for which the `species_id` is missing. Let's also
 remove observations for which `weight` and the `hindfoot_length` are missing.
 This dataset should also only contain observations of animals for which the sex
@@ -268,9 +268,9 @@ write_csv(surveys_complete, path = "data/surveys_complete.csv")
 
 ## Team Challenges
 
-These are four exercises in exploratory data analyses, which will train you to 
+These are four exercises in exploratory data analyses, which will train you to
 think about data and to use the tools you have been learning about in this class
-to solve scientific questions and to reproduce figures from the literature. To 
+to solve scientific questions and to reproduce figures from the literature. To
 solve these challenges, you will need to understand what data underlies a
 figure and how you need to manipulate it to recreate the figure.
 
@@ -371,7 +371,7 @@ surveys_abun_species %>%
     ggplot(aes(x = year, y = n, color = species)) +
         geom_line() +
         facet_wrap(~ species)
-        
+       
 # Average weight over time
 surveys_abun_species %>%
     filter(!is.na(weight)) %>%
@@ -379,7 +379,7 @@ surveys_abun_species %>%
     summarize(mean_weight = mean(weight)) %>%
     ggplot(aes(x = year, y = mean_weight, color = species)) +
         geom_line()
-        
+       
 # Total weight over time
 surveys_abun_species %>%
     filter(!is.na(weight)) %>%
@@ -418,7 +418,7 @@ under an open license and the [figures can be accessed via this
 link](http://rstb.royalsocietypublishing.org/content/368/1624/20120486.figures-only).
 
 You will be working with this dataset in assignment 3, so we have prepared the
-data for you in a format that is easier to work with. [Download the 
+data for you in a format that is easier to work with. [Download the
 data](https://raw.githubusercontent.com/UofTCoders/council/master/datasets/joel/plant-biomass-preprocess.csv)
 and read it into a dataframe called `plant_biomass`. Confirm that the data
 frame is 180 rows by 10 columns
@@ -430,16 +430,16 @@ dim(plant_biomass) == c(180, 10)
 
 Reproduce figure 3 from the paper. Focus on the overall message on the plot,
 i.e. two panels for different habitats of the plant biomass over the year in
-grazed controls vs rodent exclosures. You do not need to get the figure 
-aesthetics to look exactly the same as in the published figures (i.e. no need 
+grazed controls vs rodent exclosures. You do not need to get the figure
+aesthetics to look exactly the same as in the published figures (i.e. no need
 for the exact colors and shapes, axis styles, or to include the small dots
 around the main lines).
 
 ```{r, eval=FALSE, echo=FALSE}
 plant_biomass %>%
-    gather(species, biomass, betula_nana:vaccinium_vitis_idaea) %>% 
-    group_by(habitat, treatment, year) %>% 
-    summarize(mean_biomass = mean(biomass)) %>% 
+    gather(species, biomass, betula_nana:vaccinium_vitis_idaea) %>%
+    group_by(habitat, treatment, year) %>%
+    summarize(mean_biomass = mean(biomass)) %>%
     ggplot(aes(x = year, y = mean_biomass, color = treatment)) +
         geom_line() +
         geom_point() +
@@ -460,10 +460,10 @@ Hints:
   species (as in the paper figure).
 
 ```{r, eval=FALSE, echo=FALSE, fig.height=9, fig.width=7}
-plant_biomass %>% 
-    gather(species, biomass, betula_nana:vaccinium_vitis_idaea) %>% 
-    group_by(habitat, species, treatment, year) %>% 
-    summarize(mean_biomass = mean(biomass)) %>% 
+plant_biomass %>%
+    gather(species, biomass, betula_nana:vaccinium_vitis_idaea) %>%
+    group_by(habitat, species, treatment, year) %>%
+    summarize(mean_biomass = mean(biomass)) %>%
     ggplot(aes(x = year, y = mean_biomass, color = treatment)) +
         geom_line() +
         geom_point() +
@@ -487,10 +487,10 @@ and search online to find out how to:
 # I think are too much to ask for but can be nice to show
 color_names <- c('red', 'blue')
 
-plant_biomass %>% 
-    gather(species, biomass, betula_nana:vaccinium_vitis_idaea) %>% 
-    group_by(habitat, treatment, year) %>% 
-    summarize(mean_biomass = mean(biomass)) %>% 
+plant_biomass %>%
+    gather(species, biomass, betula_nana:vaccinium_vitis_idaea) %>%
+    group_by(habitat, treatment, year) %>%
+    summarize(mean_biomass = mean(biomass)) %>%
     ggplot(aes(x = year, y = mean_biomass, color = treatment, shape = treatment)) +
         geom_line(size = 0.8) +
         geom_point(size = 3) +
--- a/lec05-dplyr.Rmd
+++ b/lec05-dplyr.Rmd
@@ -257,15 +257,6 @@ dim(surveys_complete_comcas)
 dim(surveys_complete_naomit)
 ```
 
-```{r}
-surveys %>%
-    filter(complete.cases(species_id, weight, hindfoot_length, sex)) 
-```
-
-```{r}
-na.omit(surveys)
-```
-
 Now that our dataset is ready, we can save it as a CSV file in our `data`
 folder.
 
--- a/lec05-dplyr.Rmd
+++ b/lec05-dplyr.Rmd
@@ -9,6 +9,7 @@ title: "Tidying and exporting data, reproducing figures"
 > - Reproduce existing figures from raw data.
 > - Understand which raw data is underlying a figure.
 > - Understand which types of figures are suitable to create from raw data.
+> - Explore scientific questions using dplyr and ggplot.
 > 
 > 
 > ### Lecture outline
--- a/lec05-dplyr.Rmd
+++ b/lec05-dplyr.Rmd
@@ -214,14 +214,16 @@ Navigate to this directory in your file browser and create a folder called data 
 Alternatively, you could use R to create this directory.
 
 ```{r}
-dir.create("data_output")
+dir.create("data")
 
-# The above will work just fine, but since it gives an error when the 
-# directory already exists, you could add a a conditional expression, which only
+# To suppress the warning, we could do
+dir.create("data", showWarnings = FALSE)
+
+# Another alternative would be to use a conditional expression, which only
 # creates the directory *if* it does not already exist. The syntax here is
 # similar to the for loop we created in the second lecture.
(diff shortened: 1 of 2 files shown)


lwjohnst86:
You can just push the updated second half into this PR. No need to create a new one. In fact, creating a new one might cause some unnecessary conflicts.

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
 ---
-title: "More on using dplyr and a brief introduction to plotting with ggplot"
+title: "Tidy and export data, reproducing figures"
```
lwjohnst86: Maybe to be clearer "Tidying and exporting data"

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+    group_by(genus, plot_id) %>%
+    summarize(mean_weight = mean(weight))
+
+head(surveys_gw)
+```
+
+### Long to Wide with `spread`
+
+Now, to make this long data wide, we use `spread` from `tidyr` to spread out the
+different taxa into columns. `spread` takes three arguments - the data, the
```
lwjohnst86: `spread takes three arguments: the data, the *key* column (column with identifying information), and the *values* column (column with the numbers/values)`

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+specify what to gather than what to leave alone. And if the columns are in a
+row, we don't even need to list them all out - just use the `:` operator!
+
+```{r}
+surveys_gw_wide %>%
+  gather(genus, mean_weight, Baiomys:Spermophilus) %>%
+  head()
+```
+
+> ### Challenge
```
lwjohnst86: Previous lectures you've used `####`, but here it is three. Stick with what you've done in the past.

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+>
+> 1. Make a wide data frame with `year` as columns, `plot_id` as rows, and the
+>   values are the number of genera per plot. You will need to summarize before
+>   reshaping, and use the function `n_distinct` to get the number of unique
+>   types of a genera. It's a powerful function! See `?n_distinct` for more.
+>
+> 2. Now take that data frame, and make it long again, so each row is a unique
+>    `plot_id` `year` combination.
```
lwjohnst86:
```
`plot_id`-`year`
```

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+Before using `write_csv()`, we are going to create a new folder, `data_output`,
+in our working directory that will store this generated dataset. We don't want
+to write generated datasets in the same directory as our raw data. It's good
+practice to keep them separate. The `data` folder should only contain the raw,
+unaltered data, and should be left alone to make sure we don't delete or modify
+it. In contrast, our R code will generate the contents of the `data_output`
+directory, so even if the files it contains are deleted, we can always
+re-generate them.
+
+Use the `getwd()` function to find out which is the current working directory.
```
lwjohnst86: This may or may not start to get confusing. File systems for some reason are a challenge to learn for beginners, especially for Windows users.

This is also not super important once they start using R projects. But still, good to cover I think :)

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+```{r}
+getwd()
+```
+
+Navigate to this directory in your file browser and create a folder called data output.
+
+Alternatively, you could use R to create this directory.
+
+```{r}
+dir.create("data_output")
```
lwjohnst86: I would recommend the standard `data` name for the data folder.

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+## Exporting data
+
+Now that you have learned how to use **`dplyr`** to extract information from
+or summarize your raw data, you may want to export these new datasets to share
+them with your collaborators or for archival.
+
+Similar to the `read_csv()` function used for reading CSV files into R, there is
+a `write_csv()` function that generates CSV files from data frames.
+
+Before using `write_csv()`, we are going to create a new folder, `data_output`,
```
lwjohnst86: I'd use `data` since it is the standard name

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+in our working directory that will store this generated dataset. We don't want
+to write generated datasets in the same directory as our raw data. It's good
+practice to keep them separate. The `data` folder should only contain the raw,
+unaltered data, and should be left alone to make sure we don't delete or modify
+it. In contrast, our R code will generate the contents of the `data_output`
```
lwjohnst86: `data`

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+Alternatively, you could use R to create this directory.
+
+```{r}
+dir.create("data_output")
+
+# The above will work just fine, but since it gives an error when the 
+# directory already exists, you could add a a conditional expression, which only
+# creates the directory *if* it does not already exist. The syntax here is
+# similar to the for loop we created in the second lecture.
+if (!file.exists('data_output')) {
```
lwjohnst86: Do you mean `dir.exists`?

Also, `dir.create` has an option: `dir.create("data", showWarnings = FALSE)`

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+remove observations for which `weight` and the `hindfoot_length` are missing.
+This dataset should also only contain observations of animals for which the sex
+has been determined:
+
+```{r}
+surveys_complete <- surveys %>%
+  filter(!is.na(species_id),       # remove missing species_id
+         !is.na(weight),           # remove missing weight
+         !is.na(hindfoot_length),  # remove missing hindfoot_length
+         !is.na(sex))              # remove missing sex
```
lwjohnst86, line 245: Some alternatives are to just use:
```
suverys_complete <- na.omit(surverys)
suverys_complete <- surveys %>%
    filter(complete.cases(species_id, weight, hindfoot_length, sex)) 
    # not 100% sure this works as is. May need .$weight, etc
```

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+
+# Compare the dimensions of the original and the cleaned data frame
+dim(surveys)
+dim(surveys_complete)
+```
+
+Now that our dataset is ready, we can save it as a CSV file in our `data_output`
```
lwjohnst86: s/data_output/data

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,251 @@
...
+folder.
+
+```{r}
+write_csv(surveys_complete, path = "data_output/surveys_complete.csv")
```
lwjohnst86: s/data_output/data

lwjohnst86 (changes requested):
Very nice :)

joelostblom:
@lwjohnst86 Second half uploaded! Sorry for the delay, no worries if you don't have time to go through it. I hope the student get through 2-3 of the team challenges and added the 4th one in case anyone is really quick. If they finish early, we can have a general question period and they can take time to answer the lecture feedback survey.

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+directory, so even if the files it contains are deleted, we can always
+re-generate them.
+
+Use the `getwd()` function to find out which is the current working directory.
+
+```{r}
+getwd()
+```
+
+Navigate to this directory in your file browser and create a folder called data output.
```
lwjohnst86: `called data.`

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+folder.
+
+```{r}
+write_csv(surveys_complete, path = "data/surveys_complete.csv")
+```
+
+
+## Team Challenges
+
+These are four excersices in exploratory data analyses, where you will train to
```
lwjohnst86: `analyses, you will learn` or `analyses, which will train you to`

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+write_csv(surveys_complete, path = "data/surveys_complete.csv")
+```
+
+
+## Team Challenges
+
+These are four excersices in exploratory data analyses, where you will train to 
+think about data and to use the tools you have been learning about in this class
+to solve scientific questions and to reproduce figures from the literature. To 
+answer solve these challenges, you will need to understand what data underlies a
```
lwjohnst86: `to solve these challenges`

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+        facet_wrap(~ species)
+```
+
+If you were to look at the average hindfoot length over time, you would find 
+that the trends are similar to those of the average weight. Can you find an 
+explanation for why both the average hindfoot length and the average weight 
+decrease over time for all animals' average weight, but remain constant when 
+looking at individual species? Phrased another way: since each species display
+constant weight and hindfoot_length measures over time, what could be the cause
+to the notable decrease over time for the average weight of all species pooled
```
lwjohnst86: `what could be the cause of the notable`

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+together?
```
lwjohnst86: This is actually known as the [Simpson's Paradox](https://en.wikipedia.org/wiki/Simpson%27s_paradox)! Super cool to see in this data!

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+>    `weight`. Hint: You'll need to specify which columns are being gathered.
+>
+> 4. With this new truly long data set, calculate the average of each
+>    `measurement` in each `year` for each different `plot_type`. Then
+>    `spread` them into a wide data set with a column for `hindfoot_length` and
+>    `weight`. Hint: Remember, you only need to specify the key and value
+>    columns for `spread`.
+
+<!---
+  ```{r, echo=FALSE, purl=FALSE}
```
lwjohnst86: What's the `purl` for?

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+    filter(!is.na(weight)) %>%
+    group_by(year, species) %>%
+    summarize(mean_weight = mean(hindfoot_length)) %>%
+    ggplot(aes(x = year, y = mean_weight, color = species)) +
+        geom_line() +
+        facet_wrap(~ species)
+```
+
+
+### 2. Reproduce figure 3 from the
```
lwjohnst86: From the ... ?

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+For this section, you will apply your data wrangling and plotting skills to
+reproduce a couple of figures from a study on the yearly change in biomass of
+plants in the Abisko national park in northern Sweden. This paper is publish
+under an open license and the [figures can be accessed via this
+link](http://rstb.royalsocietypublishing.org/content/368/1624/20120486.figures-only).
+
+You will be working with this dataset in assignment 3, so we have prepared the
+data for you in a format that is easier to work with. [Download the 
+data](https://raw.githubusercontent.com/UofTCoders/council/master/datasets/joel/plant-biomass-preprocess.csv)
+and read it into a dataframe called `plant_biomass`. Confirm that you data
```
lwjohnst86: `Confirm that the data`

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+Use the `plant_biomass` data set and reproduce figure 4 from the paper.
+
+Hints:
+
+- To get the right dimensions of the subplots, explore how to use the function
+`facet_grid` instead of `facet_wrap`.
+- Remember to search online for help, e.g. "How do I change the figure size in R
+markdown?"
+- You will also need to search online to find out how you change the y-scale to
+be constant within a species, but not between species (as in the paper figure).
```
lwjohnst86: `Or you can search the help files in ?facet_grid`

lwjohnst86 (changes requested):
Really nice! I think the challenges will be fun to do!!

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+surveys <- read_csv('portal_data.csv')
+surveys
+```
+
+## Reshaping with gather and spread
+
+**`dplyr`** is one part of a larger **`tidyverse`** that enables you to work
+with data in tidy data formats. **`tidyr`** enables a wide range of
+manipulations of the structure data itself. For example, the survey data
+presented here is in almost in what we call a **long** format - every
```
QuLogic: s/is in/is

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+observation of every individual is its own row. This is an ideal format for data
+with a rich set of information per observation. It makes it difficult, however,
+to look at the relationships between measurements across plots. For example,
+what is the relationship between mean weights of different genera across the
+entire data set?
+
+To answer that question, we'd want each plot to have a single row, with all of
```
QuLogic: `plot` being land plot, right? This name is really quite unfortunate...

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+the measurements in a single plot having their own column. This is called a
+**wide** data format. For the `surveys` data as we have it right now, this is
+going to be one heck of a wide data frame! However, if we were to summarize data
+within plots and species, we might begin to have some relationships we'd want to
+examine.
+
+Let's see this in action. First, using **`dplyr`**, let's create a data frame
+with the mean body weight of each genera by plot.
```
QuLogic: s/genera/genus/

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+```
+
+Note that now the `NA` genera are included in the long format. Going from wide
+to long to wide can be a useful way to balance out a dataset so every replicate
+has the same composition.
+
+We could also have used a specification for what columns to include. This can be
+useful if you have a large number of identifying columns, and it's easier to
+specify what to gather than what to leave alone. And if the columns are in a
+row, we don't even need to list them all out - just use the `:` operator!
```
QuLogic: s/in a row/sequential/

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+  gather(genus, mean_weight, Baiomys:Spermophilus) %>%
+  head()
+```
+
+># ### Challenge
+>
+> 1. Make a wide data frame with `year` as columns, `plot_id` as rows, and the
+>   values are the number of genera per plot. You will need to summarize before
+>   reshaping, and use the function `n_distinct` to get the number of unique
+>   types of a genera. It's a powerful function! See `?n_distinct` for more.
```
QuLogic: s/a genera/a genus/

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+or summarize your raw data, you may want to export these new datasets to share
+them with your collaborators or for archival.
+
+Similar to the `read_csv()` function used for reading CSV files into R, there is
+a `write_csv()` function that generates CSV files from data frames.
+
+Before using `write_csv()`, we are going to create a new folder, `data`,
+in our working directory that will store this generated dataset. We don't want
+to write generated datasets in the same directory as our raw data. It's good
+practice to keep them separate. The `data` folder should only contain the raw,
```
QuLogic: Is `data` generated or raw? This paragraph contradicts itself.

lec05-dplyr.Rmd
QuLogic: s/exersices/exercises/

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+    ggplot(aes(x = year, y = mean_weight, color = species)) +
+        geom_line() +
+        facet_wrap(~ species)
+```
+
+If you were to look at the average hindfoot length over time, you would find 
+that the trends are similar to those of the average weight. Can you find an 
+explanation for why both the average hindfoot length and the average weight 
+decrease over time for all animals' average weight, but remain constant when 
+looking at individual species? Phrased another way: since each species display
```
QuLogic: s/display/displays/

lec05-dplyr.Rmd
QuLogic: https://youtu.be/MAMD_XpybQY

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+Reproduce figure 3 from the paper. Focus on the overall message on the plot,
+i.e. two panels for different habitats of the plant biomass over the year in
+grazed controls vs rodent exclosures. You do not need to get the figure 
+aesthetics to look exactly the same as in the published figures (i.e. no need 
+for the exact colors and shapes, axis styles, or to include the small dots
+around the main lines).
+
+```{r, eval=FALSE, echo=FALSE}
+plant_biomass %>% 
+    gather('species', 'biomass', betula_nana:vaccinium_vitis_idaea) %>%
```
QuLogic: Why are there quotes this one time?

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,510 @@
...
+- To get the right dimensions of the subplots, explore how to use the function
+`facet_grid` instead of `facet_wrap`.
+- Remember to search online for help, e.g. "How do I change the figure size in R
+markdown?"
+- You will also need to search online to find out how you change the y-scale to
+be constant within a species, but not between species (as in the paper figure).
+
+```{r, eval=FALSE, echo=FALSE, fig.height=9, fig.width=7}
+plant_biomass %>% 
+    gather('species', 'biomass', betula_nana:vaccinium_vitis_idaea) %>%
```
QuLogic: Again, why quoted?

lec05-dplyr.Rmd
joelostblom: True, should have changed it in the firs lecture with this dataset I guess...

lec05-dplyr.Rmd
joelostblom: Hahaha, they really stick to the theme ey?

lec05-dplyr.Rmd
joelostblom: Cool!! So essentially varying sample sizes as a confounding variable? I didn't know that had a name!

lec05-dplyr.Rmd
lwjohnst86: Ha! I didn't realize that was a real thing, haha wow!

lec05-dplyr.Rmd
```diff
@@ -1,4 +1,511 @@
...
+a `write_csv()` function that generates CSV files from data frames.
+
+Before using `write_csv()`, we are going to create a new folder, `data`,
+in our working directory that will store this generated dataset. We don't want
+to store manipulated datasets in the same directory as our raw data. It's good
+practice to keep them separate. The `data` folder should only contain the raw,
+unaltered data, and should be left alone to make sure we don't delete or modify
+it. In contrast, our R code will generate the contents of the `data`
+directory, so even if the files it contains are deleted, we can always
+re-generate them.
```
lwjohnst86: Yea, @QuLogic is right, this is inconsistent. I think what you mean is that the raw data should be in the `data-raw` folder (this is commonly named in R packages) and the usable/slightly modified data is in the `data` folder.

joelostblom: Wasn't it good the way it was initially? Where the output was in a separate folder?

Ideally, I would probably have both `data-raw`, and maybe `data-processed` (data here is output, but might be used as input for something else), leaving the top level clearner.

joelostblom: I imagine you will talk more in detail about these things during the project setup lecture?

lwjohnst86: Yes I will. While there are many many ways of naming folder etc in research projects, I'd prefer to stick with the R package style of doing it since it is well established and known. Plus there are tons of useful tutorials/books out there on it (e.g. http://r-pkgs.had.co.nz/).

So you would have a file structure like:

```
./
|- data/
|- data-raw/
```

lwjohnst86: data-raw would be the untouched data, and the data folder would be the 'tidied' up data

joelostblom: @lwjohnst86 Can you merge and upload to the website when you get here? Would be good to show the team challenges for the second half.

joelostblom: Actually would be good to have it in the beginning so I will merge now and we can adjust later.

PR #110, opened 2017-09-19
Draft of Lecture 6 - population models in 1D

mbonsma:
I have more material, so I'll try to time it out and see if it needs to be longer.


lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+## Lesson preamble
+
+In developing this series of lectures I have referred heavily to the notes from Nonlinear Physics,
+[PHY1460](https://www.physics.utoronto.ca/students/graduate-courses/current/phy1460s), 
+which I would encourage anyone who is interested in getting more in depth on the quantitative analysis 
+of dynamical systems to take or audit. I also refer to "Nonlinear Dynamics and Chaos" by Steven Strogatz, 
+a very good textbook which you might find helpful. You can find the pdf online, it's available as an e-book
+from the UofT library, and it's available in the Physics library as a hard copy course reserve.
+
+## Learning objectives:
```
lwjohnst86: Maybe use `###` instead, since it's under `Lesson Preamble`

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+
+- Appreciate the purpose and usefulness of modelling real-world systems
+- Understand the use of differential equations to model populations
+- Qualitatively analyze one-dimensional population models by hand and in R
+    - Find fixed points graphically
+    - Analyze stability graphically
+    - Draw one-dimensional phase portraits
+
+## Lesson outline:
```
lwjohnst86: `###`

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+of dynamical systems to take or audit. I also refer to "Nonlinear Dynamics and Chaos" by Steven Strogatz, 
+a very good textbook which you might find helpful. You can find the pdf online, it's available as an e-book
+from the UofT library, and it's available in the Physics library as a hard copy course reserve.
+
+## Learning objectives:
+
+- Appreciate the purpose and usefulness of modelling real-world systems
+- Understand the use of differential equations to model populations
+- Qualitatively analyze one-dimensional population models by hand and in R
+    - Find fixed points graphically
```
lwjohnst86: Graphically meaning by hand or graphically meaning in R?

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
 title: "Qualitative review of population models"
 ---
 
+
+## Lesson preamble
```
lwjohnst86, line 6: Maybe to be consistent with how Joel has done it, put the lesson preamble all in blockquotes

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+1. Models can simplify a complex system and help us identify what's important.
+  - biology, as we all know, is extremely complicated
+  - models summarize what we think we know about a system, 
+  force us to identify the important parameters and distill them down to a manageable number of parameters
+   
+   *Note*: that's not always how modelling is done. Sometimes you just add everything you can think of 
+   to your very complicated model, and you hope by doing that, your model will be more accurate 
+   and make better predictions. But to do that you need to simulate the model - it will likely 
+   be too complicated to solve analytically. (Analytical = a version you can leave with variables 
+   instead of putting numbers in. In R we will mostly be simulating.) An example of a discipline where
```
lwjohnst86: `simulating).`

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+   and make better predictions. But to do that you need to simulate the model - it will likely 
+   be too complicated to solve analytically. (Analytical = a version you can leave with variables 
+   instead of putting numbers in. In R we will mostly be simulating.) An example of a discipline where 
+   the models tend to be complicated is systems biology - models of celluar pathways often try to 
+   include as many components as are known, i.e. taking all the components from the following 
+   picture and creating a corresponding model.
+
+  ![Signal transduction pathway](1280px-Signal_transduction_pathways.png)
+  
+  [Image source](https://en.wikipedia.org/wiki/Modelling_biological_systems#/media/File:Signal_transduction_pathways.svg)
```
lwjohnst86: Not super important, but you could put this line into the figure caption (`Signal transduction pathway ([image source](...))`. Again, not super important.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+    - example: Lorenz model (show video)
+    - example: Lotka-volterra predator-prey model (show simulation in R)
+    - example: model from my research (show zyx vs pV)
+
+*Aside*: I do theoretical biophysics, which is quite similar to theoretical ecology in my case, 
+so all of what I'll be telling you about is techniques and concepts that I have found useful 
+or use very frequently and that I think are an important part of the theorist / data scientist toolkit. 
+
+3. In the next few lectures we will talk about models and about using R to analyze models. 
+We will first do this qualitatively - no math, then quantitatively. We will talk about fitting models to data,
```
lwjohnst86: `qualitatively (no math), then quantitatively`

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+and by the end I hope that you will be able to incorporate modelling into your project, 
+whether a model you design yourself or an existing ecological model. 
+
+## Setting up a model
+
+Let's look at some examples of one-dimensional models and talk about what we can learn from these models
+qualitatively, without doing any calculations.
+
+**Dimension**: the number of variables in your model. (Not the number of parameters -
```
lwjohnst86: ` parameters. Variables are`

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+*How many have taken some theoretical ecology before?*
+
+1. Models can simplify a complex system and help us identify what's important.
+  - biology, as we all know, is extremely complicated
+  - models summarize what we think we know about a system, 
+  force us to identify the important parameters and distill them down to a manageable number of parameters
+   
+   *Note*: that's not always how modelling is done. Sometimes you just add everything you can think of 
+   to your very complicated model, and you hope by doing that, your model will be more accurate 
+   and make better predictions. But to do that you need to simulate the model - it will likely
```
lwjohnst86: I haven't looked through the entire lecture yet, but will you be mentioning the difference between "prediction" and "inference"? Because in modelling, this can be pretty important, especially if you 'just add everything'.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+
+Let's write down a model for the growth of a species. We want to predict how many organisms are present 
+in the population at time $t$, and to do that we need to write down an equation that describes how the
+population changes with time. The notation for "change in population per change in time" is
+
+$$\frac{dN}{dt}$$
+
+If you've taken calculus, you will recognize this as the first derivative of $N$ with respect to $t$. 
+If your calculus is foggy in your memory, when you see this notation think "change in $N$ per change in $t$.
+$dN$ is the notation commonly used to mean "a small amount of $N$, so this expression mathematically describes
```
lwjohnst86: `"a small amount of $N$"` forgot the end quote

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+$\dot{N}$. I will usually use $dN/dt$ but you may see the dot notation in the literature. 
+A single dot is shorthand for "first derivative". 
+
+Biologically, $dN/dt$ is the **growth rate** or **rate of change** of a population with $N$ organisms at time $t$. 
+
+What should the growth rate of our population depend on? Let's brainstorm some things. 
+There are no wrong answers here: anything you can think of that makes biological sense is a reasonable thing 
+to include in a model, although after brainstorming we will write something very simple as an example.
+
+Possible answers:
```
lwjohnst86: Do you want to hide this section? Since most of the students are looking at the lecture notes as they are coding along during the lectures, so they will likely see this.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+- abundance of food
+
+Let's make some sample data to help us come up with a simple model.
+
+For bacteria growing in a flask with plenty of food, we know that they divide in two periodically. 
+If we take one unit of time to be the time between divisions, we can assume that the population 
+approximately doubles at each time point.
+
+```{r}
+times = seq(0, 10, by = 0.2)
```
lwjohnst86: `=` should be `<-`

lwjohnst86 (changes requested):
I haven't finished reviewing, I'll come back to it later. But looking good! Very interesting/excited to learn as well!

lec06-pop-models.Rmd
mbonsma: Both, I guess? They make the graphs in R and then find the fixed points by looking at it.

lec06-pop-models.Rmd
mbonsma: Hmm. I wasn't thinking of 'prediction' as a precise technical term here, so no, I wasn't planning on talking about inference vs. prediction.

lec06-pop-models.Rmd
mbonsma: Disagree - the period only goes outside the parentheses if the parentheses begin inside a sentence. If the entire sentence is in parentheses, the period goes inside :)

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+population_size <- 2^times
+plot(times,population_size)
```
lwjohnst86: As we discussed, `plot` could instead be `qplot` from ggplot2

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+
+$$\frac{dN}{dt} = rN$$
+
+This says that the population will increase by $r N dt$ in the small time $dt$, 
+or the population increases by $N dt$ in time $dt/r$. 
+
+The process we just used to get this model is probably not what you'd do in practice - 
+I just wanted to illustrate the connection between exponential growth and this differential equation.
+
+### Individual exercise
```
lwjohnst86: Not sure if we want to be consistent with Joel's "Challenge" sections. Up to you. I do think it should be blockquoted like Joel started it.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+qualitatively. We will draw a **phase portrait** of this system, and the techniques we use will be applicable
+to almost any model you might come across. 
+
+Let's start answering these questions qualitatively for the exponential growth model. There's a lot of
+information in the differential equation we can qualitatively read off. If we plot $dN/dt$ vs $N$, 
+the places where $dN/dt = 0$ are the fixed points of our system. Let's do this for the exponential growth model. 
+
+```{r}
+N_seq <- seq(-20,20) # make a sequence of N values to plot dN/dt against
+dN_dt = N_seq  # assume r = 1
```
lwjohnst86: s/=/<-

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+## Analyzing models (qualitatively at first)
+
+When you analyze a model, the exact questions you answer will depend on the model. 
+But in general, the following questions are ones you will try to answer for any model.
+
+1. What are the **fixed points** of the model? (For which sizes of the population(s) does the population size remain constant in time?) 
+2. Which of the fixed points are **stable**? (If the population is slightly different from the fixed point, will it go towards it or away from it?)
+3. How do the above depend on parameters of the model?
+
+---------
```
lwjohnst86: Is this hline separating content or just ending the list? If the latter, maybe just remove it?

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+
+Let's start answering these questions qualitatively for the exponential growth model. There's a lot of
+information in the differential equation we can qualitatively read off. If we plot $dN/dt$ vs $N$, 
+the places where $dN/dt = 0$ are the fixed points of our system. Let's do this for the exponential growth model. 
+
+```{r}
+N_seq <- seq(-20,20) # make a sequence of N values to plot dN/dt against
+dN_dt = N_seq  # assume r = 1
+
+plot(N_seq,dN_dt)
```
lwjohnst86: These two lines could actually be:

```
qplot(x = N_seq, y = dN_dt) +
    geom_hline(aes(yintercept = 0))
```

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+we are talking about exponential growth, after all. There's no upper limit. 
+
+--------
+
+Of course in real life there must be an upper limit to growth, otherwise the surface of the earth would be
+covered in bacteria several feet thick in a matter of hours. Let's do a quick calculation for fun - how long
+would it take for a single bacterial cell undergoing exponential growth to take over the world?
+
+```{r}
+cell_volume = 0.6 #micrometres cubed - E. coli cell volume is ~0.6 to 0.7 micrometres^3
```
lwjohnst86: s/=/<-/

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+growth_rate = 3 # let's say cells divide three times per hour - every 20 minutes (max growth rate of E. coli in good food conditions)
+
+N_t <- function(N0,r,t) { # r is growth rate - divisions per hour, t is time in hours, N0 is starting population
```
lwjohnst86: Generally, if we are going to be consistent with out own style guide, a space should come after the `,`. You can also highlight the code and type `Ctrl-Shift-A` and the code will automatically be (re-)formatted.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+  height_m = height_km*1000 # 1000 m in km
+  
+  return(height_m)
+}
+
+t<-37
+message("The bacteria would be ", round(surface_covered(2,growth_rate,t),1),  " metres thick on the surface of the earth after ", t, " hours.") #yikes. In less than two days, we would be submerged in bacteria. 
+```
+
+Let's look at another model, the *logistic growth model*, which most of you will have seen before:
```
lwjohnst86: `which some of you may have seen before`. They may not have seen it.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+Let's again plot $dN/dt$ vs. $N$ to find the fixed points:
+
+```{r}
+N_seq <- seq(-20,100) # make a sequence of N values to plot dN/dt against
+
+logistic_eqn <- function(N,r,K) { # calculate dN/dt for the logistic equation
+  return(r*N*(1-N/K))
+}
+
+dN_dt = logistic_eqn(N_seq,r=0.5,K=80)  # notice the vectorized implementation
```
lwjohnst86: s/=/<-

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+
+plot(N_seq,dN_dt)
+lines(N_seq,0*N_seq) # a line at zero for visual aid
```
lwjohnst86: As mentioned above re: `qplot`

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,298 @@
...
+```
+
+The fixed points are at $N=0$ and $N=80$. If we change $K$, we can figure out that the second fixed point is actually just $N=K$. 
+
+Now we draw the phase portrait:
+
+<!-- Phase portrait for logistic eqn -->
+
+As with the exponential growth model, the fixed point at $0$ is unstable, but now there is also a 
+fixed point at $K$ which is stable. This means that at long times, the system will go towards $N=K$.
```
lwjohnst86: `at longer times`?

lwjohnst86 (changes requested):

lec06-pop-models.Rmd
lwjohnst86: Didn't know that! Neat!

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+   *Note*: that's not always how modelling is done. Sometimes you just add everything you can think of 
+   to your very complicated model, and you hope by doing that, your model will be more accurate 
+   and make better predictions. But your model will likely 
+   be too complicated to solve analytically. (Analytical = a version you can leave with variables 
+   instead of putting numbers in. In R we will mostly be simulating.) An example of a discipline where 
+   the models tend to be complicated is systems biology - models of celluar pathways often try to 
+   include as many components as are known, i.e. taking all the components from the following 
+   picture and creating a corresponding model:  
+   
+   ![Signal transduction pathway. Image source: https://en.wikipedia.org/wiki/Modelling_biological_systems#/media/File:Signal_transduction_pathways.svg)](1280px-Signal_transduction_pathways.png)
```
QuLogic: Need to add this file.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+
+<!--*How many have taken some theoretical ecology before?*-->
+
+1. Models can simplify a complex system and help us identify what's important.
+  - Biology, as we all know, is extremely complicated.
+  - Models summarize what we think we know about a system, 
+  force us to identify the important parameters, and encourage us to distill 
+  them down to a manageable number of parameters.
+   
+   *Note*: that's not always how modelling is done. Sometimes you just add everything you can think of
```
QuLogic: Indent one more space and it should be part of the list instead of unindenting back to regular level.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+  
+2. Models can show us something we didn't expect, that we couldn't intuit from just thinking about the problem. 
+  - if your model is simple enough, it can be predictive in a broadly applicable way. We'll talk about some
+  famous ecological models, and the assignments will also include more examples. I will also include some
+  famous examples from physics, where this field is known as *nonlinear dynamics*. Many of these classic 
+  models are classic because their results weren't obvious before the model was created. 
+    - example: Lorenz model of the atmosphere which exhibits chaos ([video](https://www.youtube.com/watch?v=FYE4JKAXSfY)). I wrote a quick simulation in R which you can play around with, it's at the bottom of these notes.
+    - example: Lotka-volterra predator-prey model. 
+
+*Aside*: I do theoretical biophysics, which is quite similar to theoretical ecology in my case,
```
QuLogic: Same indent here.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+Variables are what track the species you're interested in.)
+
+Let's write down a model for the growth of a species. We want to predict how many organisms are present 
+in the population at time $t$, and to do that we need to write down an equation that describes how the
+population changes with time. The notation for "change in population per change in time" is
+
+$$\frac{dN}{dt}$$
+
+If you've taken calculus, you will recognize this as the first derivative of $N$ with respect to $t$. 
+If your calculus is foggy in your memory, when you see this notation think "change in $N$ per change in $t$.
```
QuLogic: Missing the closing double-quotes at the end.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+If we take one unit of time to be the time between divisions, we can assume that the population 
+approximately doubles at each time point.
+
+```{r}
+library(ggplot2) # so that we can use qplot
+
+times = seq(0, 10, by = 0.2)
+population_size <- 2^times
+qplot(times,population_size) +
+  geom_line(aes(x=times,y=population_size))
```
QuLogic: Spaces after comma, around `=`.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+Let's make some sample data to help us come up with a simple model.
+
+For bacteria growing in a flask with plenty of food, we know that they divide in two periodically. 
+If we take one unit of time to be the time between divisions, we can assume that the population 
+approximately doubles at each time point.
+
+```{r}
+library(ggplot2) # so that we can use qplot
+
+times = seq(0, 10, by = 0.2)
```
QuLogic: `<-`

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+qplot(times,population_size) +
+  geom_line(aes(x=times,y=population_size))
+```
+
+How do we write this assumption as a general model? Let's calculate the difference in $N$ at each time point
+and plot this vs. $N$ at the previous time point. 
+
+```{r}
+
+N_diff <- tail(population_size,-1) - head(population_size,-1) # last N-1 points minus first N-1 points
```
QuLogic: +spaces; probably put comment above to fit better.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+We will see how to answer these questions quantitatively soon. But let's start by looking at them
+qualitatively. We will draw a **phase portrait** of this system, and the techniques we use will be applicable
+to almost any model you might come across. 
+
+Let's start answering these questions qualitatively for the exponential growth model. There's a lot of
+information in the differential equation we can qualitatively read off. If we plot $dN/dt$ vs $N$, 
+the places where $dN/dt = 0$ are the fixed points of our system. Let's do this for the exponential growth model. 
+
+```{r}
+N_seq <- seq(-20,20) # make a sequence of N values to plot dN/dt against
```
QuLogic: spaces

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+dN_dt = N_seq  # assume r = 1
+
+qplot(N_seq,dN_dt)+
+  geom_line(aes(x=N_seq,y=0*N_seq)) # a line at zero for visual aid
+```
+We can also tell whether the fixed points are stable or unstable by looking at the sign of $dN/dt$
```
QuLogic: I think you need a blank line above to render a break after the plot.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+
+Of course in real life there must be an upper limit to growth, otherwise the surface of the earth would be
+covered in bacteria several feet thick in a matter of hours. Let's do a quick calculation for fun - how long
+would it take for a single bacterial cell undergoing exponential growth to take over the world?
+
+```{r}
+cell_volume = 0.6 #micrometres cubed - E. coli cell volume is ~0.6 to 0.7 micrometres^3
+growth_rate = 3 # let's say cells divide three times per hour - every 20 minutes (max growth rate of E. coli in good food conditions)
+
+N_t <- function(N0,r,t) { # r is growth rate - divisions per hour, t is time in hours, N0 is starting population
```
QuLogic: spaces; you might have to spend some time explaining functions. I'm not sure if @joelostblom went into them at all.

QuLogic: @lwjohnst86 is there a docstring-ish format we can use for these comments instead?

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+  return(N0*2^(r*t-1)) # the -1 is because the first doubling is included with the factor of 2
+}
+
+radius_earth = 6000 # approximately 6000 km 
+SA_earth = 4*pi*radius_earth^2 # surface area of earth in km^2
+
+surface_covered <- function(N0,r,t) { # calculate the volume taken up by bacteria in m on the earth's surface
+  bac_volume <- N_t(N0,growth_rate,t)*cell_volume # in micrometres cubed
+  height_km = bac_volume*(1e-27)/SA_earth # assume the earth is flat compared to the height of our surface - volume is surface area x height of surface above earth. This is height in km
```
QuLogic: Might be clearer to write that out as 1e-3^3/1e6^2 or similar instead of the final -27.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+  height_m = height_km*1000 # 1000 m in km
+  
+  return(height_m)
+}
+
+t<-37
+message("The bacteria would be ", round(surface_covered(2,growth_rate,t),1),  " metres thick on the surface of the earth after ", t, " hours.") #yikes. In less than two days, we would be submerged in bacteria. 
+```
+
+Let's look at another model, the *logistic growth model*, which most of you will have seen before:
```
QuLogic: s/most/many/ since you've already polled them...

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+and it can never be negative. But there's nothing in the model by itself, $\frac{dN}{dt} = rN$, that requires
+any of those things. Models like this are *continuous* - the variable $N$ can be any real number and isn't
+restricted to integers. 
+
+How do we reconcile this with the real world? For "large" population sizes, the continuous description is
+fairly accurate, since if $N$ is, say, 400, then a change of $\pm 1$ is a relatively small change, and we can
+think of $N$ as varying continuously. But for small populations, the continuous description breaks down. 
+If $N$ falls below $1$, the continuous description might tell you that it can bounce back, but in reality 
+$N<1$ means the population is extinct. Even though $N=0$ was an unstable fixed point for the systems we 
+looked at so far, it's what is called and **absorbing** state: once you enter it, you can never leave
```
QuLogic: s/and/an/

QuLogic (commented):
Think I got through most of this.

lec06-pop-models.Rmd
joelostblom: I was actually considering going through and removing all my blockquotes for the lessons and adding the lssons preamble heading since I noticed I missed that in some lectures...

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+   instead of putting numbers in. In R we will mostly be simulating.) An example of a discipline where 
+   the models tend to be complicated is systems biology - models of celluar pathways often try to 
+   include as many components as are known, i.e. taking all the components from the following 
+   picture and creating a corresponding model:  
+   
+   ![Signal transduction pathway. Image source: https://en.wikipedia.org/wiki/Modelling_biological_systems#/media/File:Signal_transduction_pathways.svg)](1280px-Signal_transduction_pathways.png)
+  
+2. Models can show us something we didn't expect, that we couldn't intuit from just thinking about the problem. 
+  - if your model is simple enough, it can be predictive in a broadly applicable way. We'll talk about some
+  famous ecological models, and the assignments will also include more examples. I will also include some
```
joelostblom: assignments will also include -> assignments will include

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+
+----
+
+## Introduction - why model?
+
+<!--*How many have taken some theoretical ecology before?*-->
+
+1. Models can simplify a complex system and help us identify what's important.
+  - Biology, as we all know, is extremely complicated.
+  - Models summarize what we think we know about a system,
```
joelostblom, line 45: Maybe clarify that data do not always fit theoretical models, but that it is helpful to interpret data with a larger theoretical framework in mind (as in the syllabus class learning outcomes).

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+  - Biology, as we all know, is extremely complicated.
+  - Models summarize what we think we know about a system, 
+  force us to identify the important parameters, and encourage us to distill 
+  them down to a manageable number of parameters.
+   
+   *Note*: that's not always how modelling is done. Sometimes you just add everything you can think of 
+   to your very complicated model, and you hope by doing that, your model will be more accurate 
+   and make better predictions. But your model will likely 
+   be too complicated to solve analytically. (Analytical = a version you can leave with variables 
+   instead of putting numbers in. In R we will mostly be simulating.) An example of a discipline where
```
joelostblom: will mostly be simulating -> will mostly be simulating (putting numbers in)

Maybe? Or is it overclarifying and everyone understands that already?

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+1. Models can simplify a complex system and help us identify what's important.
+  - Biology, as we all know, is extremely complicated.
+  - Models summarize what we think we know about a system, 
+  force us to identify the important parameters, and encourage us to distill 
+  them down to a manageable number of parameters.
+   
+   *Note*: that's not always how modelling is done. Sometimes you just add everything you can think of 
+   to your very complicated model, and you hope by doing that, your model will be more accurate 
+   and make better predictions. But your model will likely 
+   be too complicated to solve analytically. (Analytical = a version you can leave with variables
```
joelostblom: Nice, I really like that you enjoy these descriptions and don't assume that people know what analytical, classic, and qualitative means in these contexts. I always find myself lost when people use these and similar terms without clarifying what it means, and I feel like that hinders me from understanding.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+   include as many components as are known, i.e. taking all the components from the following 
+   picture and creating a corresponding model:  
+   
+   ![Signal transduction pathway. Image source: https://en.wikipedia.org/wiki/Modelling_biological_systems#/media/File:Signal_transduction_pathways.svg)](1280px-Signal_transduction_pathways.png)
+  
+2. Models can show us something we didn't expect, that we couldn't intuit from just thinking about the problem. 
+  - if your model is simple enough, it can be predictive in a broadly applicable way. We'll talk about some
+  famous ecological models, and the assignments will also include more examples. I will also include some
+  famous examples from physics, where this field is known as *nonlinear dynamics*. Many of these classic 
+  models are classic because their results weren't obvious before the model was created.
```
joelostblom: Created or simulated? Did people write down the equations and understood the behavior of the system by solving them analytically (and maybe the bevhavior was observed experimentally before then but no one understood it until the models were made?)? Or did people come up with the equations, but could not predict how they would develop over time by solving them analytically and it was only when they were simulated that the system behavior could be understood? Does this question make sense? =)

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+and by the end I hope that you will be able to incorporate modelling into your project, 
+whether a model you design yourself or an existing ecological model. 
+
+## Setting up a model
+
+Let's look at some examples of one-dimensional models and talk about what we can learn from these models
+qualitatively, without doing any calculations.
+
+**Dimension**: the number of variables in your model. (Not the number of parameters.  
+Variables are what track the species you're interested in.)
```
joelostblom: Maybe also clarify what model parameters are

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+- abundance of food -->
+
+Let's make some sample data to help us come up with a simple model.
+
+For bacteria growing in a flask with plenty of food, we know that they divide in two periodically. 
+If we take one unit of time to be the time between divisions, we can assume that the population 
+approximately doubles at each time point.
+
+```{r}
+library(ggplot2) # so that we can use qplot
```
joelostblom: so that we can use qplot -> which includes the qplot function

Remember that anyone not using their own computer will need to do `install.packages('ggplot2')`. You can ask them to do this before the lecture starts. 

Maybe clarify briefly that `ggplot2` is the package name, which includes several functions, including `ggplot` which we used previously, but also `qplot`. @lwjohnst86 What are the advantages using `qplot` over `ggplot`? Is it for when you don't have data frames, as indicated [here](https://stackoverflow.com/questions/5322836/choosing-between-qplot-and-ggplot-in-ggplot2)? If so, maybe mention that also @mbonsma, so they understand why you are using a different plotting function (could be helpful to mention that `qplot` is for `quickplot` for memorization).

[The qplot documentation](http://ggplot2.tidyverse.org/reference/qplot.html) mentions

> It *[qplot]* is great for allowing you to produce plots quickly, but I highly recommend learning ggplot() as it makes it easier to create complex graphics.

lec06-pop-models.Rmd
joelostblom: Or replace it with a subheading?

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+on either side of the fixed point. For a one-dimensional system, our phase portrait is a line of $N$. 
+We mark the fixed point(s) with circles, and we draw an arrow to the left wherever $dN/dt$ is negative, 
+and an arrow to the right wherever $dN/dt$ is positive in the regions between fixed points. 
+
+<!-- diagram of phase portrait --> 
+
+This model is a bit boring - there's only one fixed point, and it's unstable. This makes sense - 
+we are talking about exponential growth, after all. There's no upper limit. 
+
+--------
```
joelostblom: Same here, maybe add a subheading. It will make it easy to access the sections via the ToC

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+Let's start answering these questions qualitatively for the exponential growth model. There's a lot of
+information in the differential equation we can qualitatively read off. If we plot $dN/dt$ vs $N$, 
+the places where $dN/dt = 0$ are the fixed points of our system. Let's do this for the exponential growth model. 
+
+```{r}
+N_seq <- seq(-20,20) # make a sequence of N values to plot dN/dt against
+dN_dt = N_seq  # assume r = 1
+
+qplot(N_seq,dN_dt)+
+  geom_line(aes(x=N_seq,y=0*N_seq)) # a line at zero for visual aid
```
joelostblom: Why do you need `0 * N_seq` vs just `0`?

lec06-pop-models.Rmd
joelostblom: A few suggestions [here](https://stackoverflow.com/questions/5931292/is-there-a-sensible-way-to-do-something-like-docstrings-in-r#5934434). I am in favor of not making it too complext and just doing something like:

```{R}
N_t <- function(N0,r,t) {
    # Calculate population size changes over time
    #
    # r : growth rate as divisions per hour
    # t : time in hours
    # N0 : starting population size

  return(N0*2^(r*t-1)) # -1 since the first doubling is included with the factor of 2
}
```

or potentially [the more verbose syntax recommended in the tidyverse style guide](http://style.tidyverse.org/code-documentation.html), but I am leaning towards that it is unnecessary to spend too much time on explaining why this is necessary (unless you plan to have them write many more complex functions).

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+--------
+
+Of course in real life there must be an upper limit to growth, otherwise the surface of the earth would be
+covered in bacteria several feet thick in a matter of hours. Let's do a quick calculation for fun - how long
+would it take for a single bacterial cell undergoing exponential growth to take over the world?
+
+```{r}
+cell_volume = 0.6 #micrometres cubed - E. coli cell volume is ~0.6 to 0.7 micrometres^3
+growth_rate = 3 # let's say cells divide three times per hour - every 20 minutes (max growth rate of E. coli in good food conditions)
+
```
joelostblom: You might need to go slowly during the first function and just repeat the function syntax for them as you type it in.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+N_t <- function(N0,r,t) { # r is growth rate - divisions per hour, t is time in hours, N0 is starting population
+  return(N0*2^(r*t-1)) # the -1 is because the first doubling is included with the factor of 2
+}
+
+radius_earth = 6000 # approximately 6000 km 
+SA_earth = 4*pi*radius_earth^2 # surface area of earth in km^2
```
joelostblom: Mention that `pi` is a pre-defined variable in R, so that it is clear that we can use it without defining it ourselves.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+
+surface_covered <- function(N0,r,t) { # calculate the volume taken up by bacteria in m on the earth's surface
```
joelostblom: I would change the naming and docstring to be more precise (if I understand correctly):

```{r}
bacterial_layer_height <- function (N0, r, t) {
    # Calculate the height of the layer of evenly spread out bacteria on the earth's surface
    #
    # r : growth rate as divisions per hour
    # t : time in hours
    # N0 : starting population size
   ...
}

File: lec06-pop-models.Rmd [hunk 49]
```diff
@@ -2,3 +2,360 @@
...
+  height_km = bac_volume*(1e-27)/SA_earth # assume the earth is flat compared to the height of our surface - volume is surface area x height of surface above earth. This is height in km
+  #surface_volume = height*SA_earth
+  
+  height_m = height_km*1000 # 1000 m in km
+  
+  return(height_m)
+}
+
+t<-37
+message("The bacteria would be ", round(surface_covered(2,growth_rate,t),1),  " metres thick on the surface of the earth after ", t, " hours.") #yikes. In less than two days, we would be submerged in bacteria.
```
joelostblom: I think it would be a good idea to graph this instead of outputting a message with just one value:

```{r}
t <- seq(0, 38, 0.3)
qplot(t, bacterial_layer_height(2, growth_rate, t))
```

I was going to suggest to have even more points and change the point size, but turns out this is rather unintuitive for `qplot` (`size=I(0.2)`), so might just skip it.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+Let's again plot $dN/dt$ vs. $N$ to find the fixed points:
+
+```{r}
+N_seq <- seq(-20,100) # make a sequence of N values to plot dN/dt against
+
+logistic_eqn <- function(N,r,K) { # calculate dN/dt for the logistic equation
+  return(r*N*(1-N/K))
+}
+
+dN_dt = logistic_eqn(N_seq,r=0.5,K=80)  # notice the vectorized implementation
```
joelostblom: I only had time to touch briefly on for loops during the lecture in class, so maybe just clarify that vectorized here refers to that R performs matrix algebra faster under the hood, than if we would loop through each element manually within a for loop. And the code also become easier to read with the current notation.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+As with the exponential growth model, the fixed point at $0$ is unstable, but now there is also a 
+fixed point at $K$ which is stable. This means that at long times, the system will go towards $N=K$. 
+
+### Note: the importance of stability
+
+Why is stability important? The real world is noisy, and while mathematically an unstable fixed point is 
+still a fixed point (if you were exactly there you would stay there forever), in reality there will 
+always be small perturbations (perhaps a member of the population dies or migrates in) that will move 
+you away from the fixed point. The important thing to know, then, is whether the population will remain 
+close to that fixed point or move away to a different point.
```
joelostblom, line 339: Nice to contextualize with real world examples and mentioning perturbations! Actually helps me to think about this =)

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+Why is stability important? The real world is noisy, and while mathematically an unstable fixed point is 
+still a fixed point (if you were exactly there you would stay there forever), in reality there will 
+always be small perturbations (perhaps a member of the population dies or migrates in) that will move 
+you away from the fixed point. The important thing to know, then, is whether the population will remain 
+close to that fixed point or move away to a different point. 
+
+Think of a ball balanced perfectly on top of a hill, or think of balancing a pen on your finger, 
+or think of balancing a chair on the back two legs. It can stay there as long as there are no disturbances, 
+but the tiniest breath of wind or tremble in your balance will knock it over and it will go towards its other
+fixed point - the ground.
```
joelostblom, line 344: From you last paragraph, I thought of unstable vs stable fixed points as a ball on top of a hill vs a ball in a (endless?) bowl/valley. Both are stable, but the surrounding landscape either amplifies or dampens the forces of perturbations.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+
+### Note: discrete vs. continuous
+
+There is a subtlety to these differential equation descriptions of models. To see this, calculate
```
joelostblom: Add a `#### Challenge` heading for this.

lec06-pop-models.Rmd
```diff
@@ -2,3 +2,360 @@
...
+
+print(r*N)
+```
+
+Clearly, if $N$ represents a real population, it can only ever take integer values (it must be *discrete*), 
+and it can never be negative. But there's nothing in the model by itself, $\frac{dN}{dt} = rN$, that requires
+any of those things. Models like this are *continuous* - the variable $N$ can be any real number and isn't
+restricted to integers. 
+
+How do we reconcile this with the real world? For "large" population sizes, the continuous description is
```
joelostblom, line 366: Again, great that you explain how this relates to the real world!

joelostblom (commented):
Great job! I already learnt a lot and I am looking forward to the lecture! Do you think it would be possible to include a shorter challenge activity about half way into the lecture? To get them thinking and trying out `qplot` themselves, something very similar to what you did but with difference numbers and have them understand where the fixed point is etc.

lec06-pop-models.Rmd
mbonsma: Not sure what this means?

QuLogic: search & replace (on this line)

lec06-pop-models.Rmd
mbonsma: Both! The two examples I have at the beginning are examples of one that was simulated (Lorenz) and one that is typically understood analytically (Lotka-Volterra).

lec06-pop-models.Rmd
mbonsma: Yeah. Mostly because the plots aren't, like, *nice*, I just wanted something quick.

lec06-pop-models.Rmd
mbonsma: Neat! I just assumed it had to be a length(N_seq) vector, but that works!

lec06-pop-models.Rmd
mbonsma: The top of the hill isn't stable though. It's unstable because if you move away from the top of the hill, you won't go back towards it.

QuLogic: @joelostblom what you're thinking of is that both are in *equilibrium*.

joelostblom: Sorry I should have said "Both are *fixed*" instead of "Both are stable". I just meant that if there are no forces acting upon the ball, it will stay on top of the hilll (give that it is exactly at the top of the hill). So that very point is fixed as you said, but wether it is stable or unstable has to do with the properties of the landscape, not the point. This is really just me trying to understand the concept by typing it in different words.

lec06-pop-models.Rmd
joelostblom: and -> an

I still like that syntax more, although it does not give as many geekpoints

lec06-pop-models.Rmd
joelostblom: What @QuLogic said sounds good, that's probably the term I should have used: "Both are in equilibrium"

mbonsma: @joelostblom yes exactly! Both are *fixed*: if you are exactly on top of it, you will stay there. It's the curvature of the landscape near the fixed point that determines stability, which is why, as we will get to next lecture maybe, you calculate derivatives to assess stability analytically.

mbonsma:
Thanks for your comments everyone! I'm excited for tomorrow!

joelostblom (approved):
Looks great!

lwjohnst86 (approved):

PR #116, opened 2017-09-21
minor (unimportant) typo

lwjohnst86:


lwjohnst86:
one more item coming up

PR #124, opened 2017-09-27
Lecture 7 - modelling in R

mbonsma:
I'm done up to line 449 - 'Qualitative analysis of two-dimensional population models'

Feedback I want specifically:
- does this go too in depth into numerically solving equations in R? 
- line [440](https://github.com/UofTCoders/rcourse/compare/master...mbonsma:lec7?expand=1#diff-b557fa26912219c7adcb8f6bb35f78e1R440) - I'm trying to use ggplot, but I don't know how to best do it for this example. Also on line [428](https://github.com/UofTCoders/rcourse/compare/master...mbonsma:lec7?expand=1#diff-b557fa26912219c7adcb8f6bb35f78e1R428), I want to do a `full_join` on all the dataframes in that list, but I don't know how to do it for the entire list. Any suggestions?


lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
 ---
-title: "Qualitative population models in 2 dimensions"
+title: "Population models in two dimensions and numerically solving equations in R"
+author: "Madeleine Bonsma-Fisher"
+output: html_document
```
lwjohnst86: You don't need author and output. Though, it probably would be a good idea to list who created which lessons.... Keep author in!

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+> ### Lesson outline
+> 
+> Total lesson time: 2 hours
+> 
+> - 
+
+> ### Setup
+> 
+> - `install.packages('deSolve')`
```
lwjohnst86: Because you are using a new package, you can do one of two things in order for it to work on the website:

- Run `devtools::use_package("deSolve")` in the console (install devtools if you don't have it)
- Open up DESCRIPTION file and add `deSolve` to the line right below `Imports:`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+## Recap and some tips
+
+   - You can include math in R notebooks using [LaTeX syntax](https://en.wikibooks.org/wiki/LaTeX/Mathematics). 
+
+     `$$\frac{dN}{dt} = r N (1 - N/K)$$` produces 
+     $$\frac{dN}{dt} = r N (1 - N/K)$$
+
+   - **Differential equation**: an equation that describes how a function of variables is related to
+derivatives of those variables.
```
lwjohnst86: tab this line to match previous line, since it is part of the item

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+     `$$\frac{dN}{dt} = r N (1 - N/K)$$` produces 
+     $$\frac{dN}{dt} = r N (1 - N/K)$$
+
+   - **Differential equation**: an equation that describes how a function of variables is related to
+derivatives of those variables. 
+
+   - Drawing **phase portraits** in one dimension:
+
+    ![Logistic equation simulation](https://github.com/UofTCoders/rcourse/raw/master/logistic.gif)
```
lwjohnst86: Are you creating this logistic.gif file? Or is it from the internet? If it is, maybe just link directly to the website.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+       - Fixed points: values of $N$ at which $dN/dt$, the rate of change of $N$, is $0$. 
+       To find fixed points, plot $dN/dt$ vs. $N$ and find the place(s) where it crosses the $x$ axis. 
+       - Stability: if you start at some $N$ close to the fixed point but not exactly on it, 
+       will you go towards (stable) or away (unstable) from the fixed point? The sign of $dN/dt$ on
+       either side of a fixed point tells you whether $N$ will increase or decrease in that area. 
+       Draw an arrow to the right if $dN/dt$ is positive, and draw and arrow to the left if $dN/dt$ is negative. 
+
+### Challenge
+
+Find the fixed point(s) of the following differential equation.
```
lwjohnst86: Which equation? I assume you'll add it eventually. Ah, I am guessing the equation is the one on line 53?

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+   - Plot $\frac{dN}{dt}$ vs. $N$ for $N$ ranging from $0$ to $10$.
+   - Sketch a phase portrait. Are the fixed point(s) stable or unstable?
+
+$$\frac{dN}{dt} = \text{sin}(N)$$
+<!--
```
lwjohnst86: Instead of wrapping in `<!--` you could do this instead:

```
```{r, echo=FALSE, eval=FALSE}
```

This will hide the code.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+```{r}
+library(ggplot2)
+
+N_seq = seq(0, 10, by = 0.1)
+
+dN_dt = sin(N_seq)
+
+qplot(N_seq, dN_dt) +
+  geom_line(aes(x = N_seq, y = 0))
```
lwjohnst86: Instead of using `geom_line` you could simplify this by using:

```
geom_hline(aes(yintercept = 0))
```

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+   - Plot $\frac{dN}{dt}$ vs. $N$ for $N$ ranging from $0$ to $10$.
+   - Sketch a phase portrait. Are the fixed point(s) stable or unstable?
+
+$$\frac{dN}{dt} = \text{sin}(N)$$
+<!--
+```{r}
+library(ggplot2)
+
+N_seq = seq(0, 10, by = 0.1)
```
lwjohnst86: s/=/<-

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+    ![Logistic equation simulation](https://github.com/UofTCoders/rcourse/raw/master/logistic.gif)
+
+       - Fixed points: values of $N$ at which $dN/dt$, the rate of change of $N$, is $0$. 
+       To find fixed points, plot $dN/dt$ vs. $N$ and find the place(s) where it crosses the $x$ axis. 
+       - Stability: if you start at some $N$ close to the fixed point but not exactly on it, 
+       will you go towards (stable) or away (unstable) from the fixed point? The sign of $dN/dt$ on
+       either side of a fixed point tells you whether $N$ will increase or decrease in that area. 
+       Draw an arrow to the right if $dN/dt$ is positive, and draw and arrow to the left if $dN/dt$ is negative. 
+
+### Challenge
```
lwjohnst86: s/###/#### (four headers)

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+  geom_line(aes(x = N_seq, y = 0))
+```
+-->
+
+You can think of $dN/dt$ as a *velocity*: it tells you how fast $N$ is changing and in what
+direction. If we start a **trajectory** at some initial $N_0$, $dN/dt$ dictates whether $N$ will
+increase, decrease, or stay the same. If $N_0$ happens to be at a fixed point, $N$ will remain at
+$N_0$ for all time. 
+
+### Challenge (time permitting)
```
lwjohnst86: `####` (four header)

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+
+Find the fixed point(s) of the following differential equation.
+
+   - Plot $\frac{dx}{dt}$ vs. $x$ for $x$ ranging from $-2$ to $2$ in $0.1$ increments.
+   - Sketch a phase portrait. Are the fixed point(s) stable or unstable?
+
+$$\frac{dx}{dt} = x^2 - 1$$
```
lwjohnst86: I'd suggest this equation be moved to below where is says "of the following equation" (line 75)

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+<!--
+```{r}
```
lwjohnst86: See above re: `echo=FALSE, eval=FALSE`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+library(ggplot2)
+
+x_seq = seq(-2, 2, by = 0.1)
```
lwjohnst86: s/=/<-

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+dx_dt = x_seq^2 - 1
+
+qplot(x_seq, dx_dt) +
+  geom_line(aes(x = x_seq, y = 0))
```
lwjohnst86: See above about using `geom_hline` (horizontal line)

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+```
+-->
+
+
+## Numerically simulating models in R
+
+I'm moving this topic ahead so that we can use what we learn here to aid us in analyzing more
+complicated models in one and two dimensions. 
+
+Some differential equations are possible to solve analytically (by integration). But when we encounter equations that are hard to solve analytically, we can instead solve them **numerically**. Numerically solving a differential equation means calculating an approximate solution for the variable of interest as it evolves in time, using the information contained in the equation about the *rate of change* of the variable at each time point.
```
lwjohnst86: Hard wrap

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+qplot(x_seq, dx_dt) +
+  geom_line(aes(x = x_seq, y = 0))
+```
+-->
+
+
+## Numerically simulating models in R
+
+I'm moving this topic ahead so that we can use what we learn here to aid us in analyzing more
+complicated models in one and two dimensions.
```
lwjohnst86: Is this text needed?

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+  # r: growth rate (divisions per hour) 
+  # K: carrying capacity
+  # N: population size
+  return(r * N * (1 - N / K))
+  }
+  
+dN_dt <- logistic_eqn(N_seq, r = 0.5, K = 80)
+  
+qplot(N_seq, dN_dt) +
+  geom_line(aes(x = N_seq, y = 0)) # a line at zero for visual aid
```
lwjohnst86: `geom_hline`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+The image below is a cartoon of what this looks like in practice. Starting from a point $A_0$, we use
+the derivative to tell us in what direction we should take our next step. $\Delta t$ is a parameter
+we can choose that determines how large of a step we take. 
+
+This process has several names: **Euler's method** after Leonhard Euler who wrote about it 
+in about 1770, or **forward finite difference method**, referring to the process of stepping 
+forward in time in small (*finite*) increments (*difference*). 
+
+![Euler's method: blue is the true solution, red is the approximate solution. Image source: https://commons.wikimedia.org/w/index.php?curid=2143753](412px-Euler_method.png)
```
lwjohnst86: Instead of uploading the png to the repo, you could just link directly to the website.

```
![Euler's method: blue is the true solution, red is the approximate solution. Image source: Wikimedia](https://upload.wikimedia.org/wikipedia/commons/thumb/1/10/Euler_method.svg/412px-Euler_method.svg.png)
```

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+There are many related but slightly different techniques for numerically solving differential
+equations. The best method will often depend on the situation, and we won't go into detail on any
+other methods, but you can look up *Runge-Kutta* methods if you're interested in learning more about it.
```
lwjohnst86: Maybe provide a link here for the Runge-Kutta method for them to look at if they want.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+If we put in $N_0$ to $dN/dt$:
+
+$$N_1 = rN_0(1-\frac{N_0}{K}) + N_0$$
+
+Now we have a thing we can calculate! Plugging in the numbers we chose for $r$, $K$, and $N_0$, we get
+
+$$N_1 = 1 \times 2(1-\frac{2}{10}) + 2$$
+$$N_1 =  2+ 1.6 = 3.6$$
+We repeat the proccess to get $N_2$, and so on, remembering to evaluate $dN/dt$ at the value of $N$ from the previous timestep. The vertical bar $\vert$ instructs us to substitute $N_0$ for $N$ in the derivative.
```
lwjohnst86: Hard wrap

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+more and more negative after that. 
+
+## Using `deSolve` in R to solve differential equations numerically
+
+R has a package called `deSolve` which contains several functions to numerically solve differential equations.
+Let's look at the documentation for `deSolve`:
+
+```{r}
+library(deSolve)
+help(deSolve)
```
lwjohnst86: This command shouldn't be used in a rmarkdown document. You can put the code chunk so it doesn't evaluate:

    ```{r}
    library(deSolve)
    ```
    ```{r, eval=FALSE}
    help(deSolve)
    ```

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+This is pretty hard to read, especially if your calculus was a long time ago. The important thing to know for
+our purposes is that the equations we're working with are **first-order ordinary differential equations**.
+*First-order* means that the derivatives are all first derivatives (as opposed to second derivatives, third
+derivatives, etc.). **Ordinary** means that the derivatives are full derivatives (as opposed to partial
+derivatives which are written $\partial N / \partial t$). 
+
+This means that the function we'll be using is called `ode`, which stands for 'ordinary differential equation'. 
+
+```{r}
+help(ode)
```
lwjohnst86: Same as above. Use `eval=FALSE`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+```{r}
+library(deSolve)
+help(deSolve)
+```
+
+This is pretty hard to read, especially if your calculus was a long time ago. The important thing to know for
+our purposes is that the equations we're working with are **first-order ordinary differential equations**.
+*First-order* means that the derivatives are all first derivatives (as opposed to second derivatives, third
+derivatives, etc.). **Ordinary** means that the derivatives are full derivatives (as opposed to partial
+derivatives which are written $\partial N / \partial t$).
```
lwjohnst86, line 374: I don't understand this. What is the difference between full and partial besides the different math symbol?

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+```
+
+The method used by `ode` to numerically solve differential equations is called `lsoda`, which automatically
+tries to use the best method depending on the equation. This is why people like to use functions like `ode`:
+the result is often more accurate than Euler's method and might also be faster to run. 
+
+```{r}
+# Example: deSolve with logistic eqn
+
+library(deSolve)
```
lwjohnst86: You already did this above

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+# now we need to define our function a bit differently to be in the format that `ode` uses
+logistic_fn <- function(t,state,parameters) {
```
lwjohnst86: Why use only three args? Is this an `ode` thing?

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+  K <- parameters[2] # the second element of the parameters vector is K
+
+  #rate of change
+  dN = r*N*(1-N/K)
+    
+  #return rate of change
+  list(c(dN))
+}
+
+parameters <- c(r=0.5,K=50)
```
lwjohnst86: spaces around `=`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+  # state: vector of variables (here it's just N)
+  # parameters: vector of parameters c(r, K)
+  
+  N <- state 
+  
+  r <- parameters[1] # the first element of the parameters vector is r
+  K <- parameters[2] # the second element of the parameters vector is K
+
+  #rate of change
+  dN = r*N*(1-N/K)
```
lwjohnst86: s/=/<-

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+    
+  #return rate of change
+  list(c(dN))
+}
+
+parameters <- c(r=0.5,K=50)
+state <- c(N=10)
+times <- seq(0, 50, by = 0.01)
+
+help(ode) # look at the documentation to learn about the parameters
```
lwjohnst86: Comment this out, so it doesn't get added to the html output

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+result <- ode(y = state, times = times, func = logistic_fn, parms = parameters)
+```
+
+The output of `ode` is a vector that contains the times we requested and its calculated values of N. 
+
+Notice that by using `ode` we didn't have to write a loop explicitly ourselves, and we didn't have to
+choose a timestep. `ode` does under the hood a version of what we did explicitly before. 
+
+```{r}
+print(head(result,10)) # result has a list of times and values of N at those times
```
lwjohnst86: space after `,`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+print(typeof(result)) # result is a 'double': a vector of numbers.
```
lwjohnst86: Joel did not go over `typeof`. He used `class`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+ggplot(result) +
+  geom_point(aes(x = time, y = N))
+```
+
+Now let's make a nice plot of numerical solutions from a few different starting points.
+
+```{r}
+library(dplyr)
+
+parameters <- c(r=0.5,K=50)
```
lwjohnst86: spaces

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+  
+  result <- ode(y = N0, times = times, func = logistic_fn, parms = parameters)
+  result <- data.frame(result)
+  
+  results[[count]] <- result
+  
+  count <- count + 1
+}
+
+all_results <- bind_cols(results)
```
lwjohnst86: I'm going to submit a PR to your fork of some code suggestions for this. It will be way easier than doing it here.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+```
+
+```{r}
+View(all_results) # look at what we got
```
lwjohnst86: Shouldn't use `View` if knitting

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+print(colnames(all_results)) # column names
+```
+
+
+
+```{r}
+ggplot(all_results) +
```
lwjohnst86: Same as above, I'll submit a PR to your fork

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+ - introduce isoclines? null cline is a special case of isocline
+
+
+
+```{r}
+library(deSolve)
+lotka_volterra <- function(t,y,parameters) { 
+  # the arguments need to be named "t", "y", and "parameters", otherwise this won't work with phaseR
+    
+    X <- y[1]; Y <- y[2]
```
lwjohnst86: Generally this is not recommended. Each command should be a new line

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+    dx <- alpha*X - beta*X*Y
+    dy <- delta*X*Y - gamma*Y
+    
+    #return rate of change
+    list(c(dx, dy))
+}
+```
+
+```{r}
+lotka_volterra(2,y = c(0.6,0.5), parameters = c(0.5,0.6,0.7,0.8))
```
lwjohnst86: space after `,`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+```
+
+
+```{r}
+# deSolve with lotka volterra
+
+#parameters = c(alpha = 0.5, beta = 0.6, gamma = 0.5, delta = 0.5)
+
+parameters = c(0.5, 0.6, 0.9, 0.5)
```
lwjohnst86: s/=/<-

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+state <- c(X = 0.5,
+ Y = 0.5) # these variables need to match the function arguments in label
```
lwjohnst86: Why new line here?

lwjohnst86: I'm stopping at this point since I am guessing you are still working on this bit.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,617 @@
...
+
+dN <- logistic_eqn(N1, r, K)*dt # calculate what the next change in N would be
+
+N2 <- N1 + dN
+
+print(N2) # N is getting more and more negative!
+ 
+```
+
+### Challenge - in groups
```
lwjohnst86: `####`

lwjohnst86 (changes requested):
Great job overall! I'm going to submit a PR to your fork with my code suggestions for the items you specifically wanted feedback on. 

Whether this is going into too much depth for numerically solving equations... I don't know. I mean, this was first and foremost a "Theoretical Ecology" course that the students enrolled in, so I would assume they would appreciate it. Now, whether this stuff will be useful for their final project ... I don't know. What do you think, will it be useful?

mbonsma:
I mean, if they have to anything with models ever this is a good thing to know. I suppose we have an entire lecture allocated for it anyway, so it's not like I'm spending more time on it than I planned. It's just not in the original order, that's all.

lec07-pop-models.Rmd
mbonsma: It's been added to the PR now. I made it :)

lec07-pop-models.Rmd
mbonsma: There's no difference when it's a function of only one variable (like time), but if you had something that depended on two things, like position (x) and time (t), then partial derivative would be taking the derivative only with respect to one of those variables. A full derivative means you have to chain rule the shit out of everything to catch anything that implicitly depends on the variable.  

Example: y(x,t) = xt. If you took the full derivative with respect to time, say, you'd have to also differentiate x with respect to time. dy(x,t)/dt = t dx/dt + x, whereas partial y / partial t = x.

mbonsma: It doesn't matter for this course, I just didn't want them to be scared or confused when looking at the help document.

lec07-pop-models.Rmd
mbonsma: oookkaaaayy

whatev

why can't I import things a lot

lec07-pop-models.Rmd
mbonsma: It's an `ode` thing :( it has this very specific argument format, as far as I can tell. `state` and `parameters` have to be vectors.

lec07-pop-models.Rmd
mbonsma: gah I can't stop doing this

lec07-pop-models.Rmd
mbonsma: oh ok that's what I wanted then.

lec07-pop-models.Rmd
mbonsma: Yep this part isn't done yet.

lec07-pop-models.Rmd
lwjohnst86: I forgot to say you also need to add `phaseR` to the `Imports:`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+>
+> - Review differential equations and one-dimensional phase portraits
+> - Numerically solve differential equations in R using Euler's method and R's ODE solver
+> - Qualitatively analyze two-dimensional models by hand and in R
+
+> ### Lesson outline
+> 
+> Total lesson time: 2 hours
+> 
+> -
```
joelostblom: Don't forget to add this

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+   - **Differential equation**: an equation that describes how a function of variables is related to
+   derivatives of those variables. 
+
+   - Drawing **phase portraits** in one dimension:
+
+    ![Logistic equation simulation](https://github.com/UofTCoders/rcourse/raw/master/logistic.gif)
+
+       - Fixed points: values of $N$ at which $dN/dt$, the rate of change of $N$, is $0$. 
+       To find fixed points, plot $dN/dt$ vs. $N$ and find the place(s) where it crosses the $x$ axis. 
+       - Stability: if you start at some $N$ close to the fixed point but not exactly on it,
```
joelostblom: Can it goes towards from one direction and then away from the other side of the fixed point?

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+direction. If we start a **trajectory** at some initial $N_0$, $dN/dt$ dictates whether $N$ will
+increase, decrease, or stay the same. If $N_0$ happens to be at a fixed point, $N$ will remain at
+$N_0$ for all time. 
+
+#### Challenge (time permitting)
+
+
+Find the fixed point(s) of the following differential equation.
+
+$$\frac{dx}{dt} = x^2 - 1$$
```
joelostblom: Why do you use `x` instead of `N` here? Is it to say that it could be any equation and not only the number of individuals?

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+  }
+  
+dN_dt <- logistic_eqn(N_seq, r = 0.5, K = 80)
+  
+qplot(N_seq, dN_dt) +
+  geom_hline(aes(yintercept = 0)) # a line at zero for visual aid
+```
+
+$dN/dt$ is the **slope** of $N$ with respect to $t$ at a given value of $N$. 
+We can use the slope as an update rule, exactly like the recursion equation we
```
joelostblom, line 148: What does "update rule" mean? That at the smallest possible step, the model will update with the slope from the previous step?

joelostblom: Ok I understand this after reading further

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+  # calculate dN/dt for the logistic equation
+  
+  # r: growth rate (divisions per hour) 
+  # K: carrying capacity
+  # N: population size
+  return(r * N * (1 - N / K))
+  }
+
+# parameters
+K <- 150; r <- 1  # you can put things on the same line, separated by a semicolon
```
joelostblom: We don't really want them to do this though. It's can become messy quickly and there is no real advantage to it, at least not in our class.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+
+dt <- 0.05 #timestep - the smaller, the better
```
joelostblom: space after `#`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+t_vector <- seq(0, tmax - dt, by = dt) # time vector
+
+# initial condition
+N0 <- 10
+N_vector[1] <- N0 
+
+N <- N0 # initialize variable N
+
+for (i in 2:points){ # start at 2 because the initial state is at position 1
+  dN <- logistic_eqn(N = N, r = r, K = K)*dt
```
joelostblom: spaces around multiplication operator

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+N1 <- N0 + dN
+
+print(N1) # this is no good: N has shot from +400 to -1000 in a single step! 
+
+dN <- logistic_eqn(N1, r, K)*dt # calculate what the next change in N would be
+
+N2 <- N1 + dN
+
+print(N2) # N is getting more and more negative!
+
```
joelostblom, line 338: Nice, maybe add a short note about the drawbacks of setting `dt` too small as well. I imagine that these mainly are computational cost and time to solve the equation. Is there anything analogous to gradient descent where choosing a too small update step can get you stuck in a local optima and never find the best global solution to a problem?

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+dN <- logistic_eqn(N1, r, K)*dt # calculate what the next change in N would be
+
+N2 <- N1 + dN
+
+print(N2) # N is getting more and more negative!
+ 
+```
+
+#### Challenge - in groups
```
joelostblom, line 341: Make sure they sit in close to eachother from the beginning and no one ends up at an isolated desktop computer in the second row.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+## Using `deSolve` in R to solve differential equations numerically
+
+R has a package called `deSolve` which contains several functions to numerically solve differential equations.
+Let's look at the documentation for `deSolve`:
+
+```{r}
+library(deSolve)
+```
+```{r, eval=FALSE}
+help(deSolve)
```
joelostblom: We have done `?deSolve` so far, so maybe just mention that these give the same results and`?` is just a shortcut for `help()`

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+  # state: vector of variables (here it's just N)
+  # parameters: vector of parameters c(r, K)
+  
+  N <- state 
+  
+  r <- parameters[1] # the first element of the parameters vector is r
+  K <- parameters[2] # the second element of the parameters vector is K
+
+  #rate of change
+  dN <- r*N*(1-N/K)
```
joelostblom: spaces

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+times <- seq(0, 50, by = 0.01)
+
+#help(ode) # look at the documentation to learn about the parameters
+result <- ode(y = state, times = times, func = logistic_fn, parms = parameters)
+```
+
+The output of `ode` is a vector that contains the times we requested and its calculated values of N. 
+
+Notice that by using `ode` we didn't have to write a loop explicitly ourselves, and we didn't have to
+choose a timestep. `ode` does under the hood a version of what we did explicitly before.
```
joelostblom: Change

> `ode` does under the hood a version of what we did explicitly before.

to

> Under the hood, `ode` performs a similar computation to what we did manually. The `ode` version of this calculation is faster than our implementation, so we save both time writing and executing the code.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+
+$$\frac{dx}{dt} = ax - bxy$$
+$$\frac{dy}{dt} = cx - dy$$
+
+This is sort of like the Lotka-Volterra predator-prey model, but the term for predator growth is $cx$ instead of $cxy$. In this model the predators will eat the prey at a constant rate, no matter how many of the predators there are. 
+
+First, we'll simulate a trajectory for the two species.
+
+```{r}
+lotka_volterra <- function(t,y,parameters) {
```
joelostblom: spaces

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+  geom_line(aes(x = time, y = X), color = 'red') +
+  geom_line(aes(x = time, y = Y), color = 'blue') 
+```
+
+It looks like we have oscillations that decay to a fixed point. To find out more generally what the fixed points are, we can plot 
+
+ - fixed points
+ - stability
+ - phase portraits
+ - introduce null clines - *cline* comes from a Greek word meaning "slope"
```
joelostblom: Cool, I love learning about the origins of words haha!

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,657 @@
...
+qplot(N_seq, dN_dt) +
+  geom_hline(aes(yintercept = 0))
+```
+
+You can think of $dN/dt$ as a *velocity*: it tells you how fast $N$ is changing and in what
+direction. If we start a **trajectory** at some initial $N_0$, $dN/dt$ dictates whether $N$ will
+increase, decrease, or stay the same. If $N_0$ happens to be at a fixed point, $N$ will remain at
+$N_0$ for all time. 
+
+#### Challenge (time permitting)
```
joelostblom, line 94: I do not think you will have time for this. You have *plenty* of material already (which is great!).

joelostblom (commented):
Nice, great work!! It is really quite educational to read through these pull requests, I like it =) There is so much stuff in here that I think you might have to move some over to next class, but it depends on how fast the class gets the concepts (they might not be as slow as I am...). Just make sure you give them enough time to work on the challenges and move over things to next class if you don't get through everything.

joelostblom:
Oh and regarding the specific items: I didn't feel like it was too much depth for numerical methods reading through it (and I didn't even have this question in mind cause I only saw your comment now, so it is an unbiased observation).  I will let Luke give feedback on your R questions, cause I don't know.

lec07-pop-models.Rmd
mbonsma: I'll do that for next lecture - I don't think I'll get to it tomorrow.

mbonsma: To using phaseR, I mean.

lec07-pop-models.Rmd
mbonsma: Yeah, in a 1D system you could have "half-stable" where it's stable from one direction but unstable on the other side. 

In 2D systems, you can have the flow going towards a fixed point in one eigendirection and away in another. It's still technically unstable, but it has a special name: **saddle node**, because if you imagine the stability landscape as a hill, that fixed point is like a mountain pass.

lec07-pop-models.Rmd
mbonsma: Yeah exactly. I expect that changing to $x$ will throw them off, and I want them to realize that you can name the variables whatever you want.

lec07-pop-models.Rmd
mbonsma: It's not necessarily faster to execute, but this isn't something I know a lot about.

lec07-pop-models.Rmd
joelostblom: Makes sense, thanks for the explanation!

lec07-pop-models.Rmd
lwjohnst86: PS: submitted a [PR to your fork](https://github.com/mbonsma/rcourse/pull/2). Let me know if you want me to explain any of it. `do` just applies a function to each initial condition in the data.frame, like the loop does, but in the tidyverse style of doing it. So that the output is a data frame, which you can than continue in a pipe for whatever else you want to do.

lwjohnst86: If you merge that PR into your fork, it will update automatically here.

lec07-pop-models.Rmd
QuLogic: Why `devtools` instead of regular `install.packages` (which seems to have worked)?

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,591 @@
...
+   - Drawing **phase portraits** in one dimension:
+
+    ![Logistic equation simulation](https://github.com/UofTCoders/rcourse/raw/master/logistic.gif)
+
+       - Fixed points: values of $N$ at which $dN/dt$, the rate of change of $N$, is $0$. 
+       To find fixed points, plot $dN/dt$ vs. $N$ and find the place(s) where it crosses the $x$ axis. 
+       - Stability: if you start at some $N$ close to the fixed point but not exactly on it, 
+       will you go towards (stable) or away (unstable) from the fixed point? The sign of $dN/dt$ on
+       either side of a fixed point tells you whether $N$ will increase or decrease in that area. 
+       Draw an arrow to the right if $dN/dt$ is positive, and draw and arrow to the left if $dN/dt$ is negative.
```
QuLogic: draw and arrow -> draw an arrow

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,591 @@
...
+Let's choose $r = 1$, $K = 10$, $N_0 = 2$, and $\Delta t = 1$ to make our math easier.
+
+Now we have $N_{t=0} = N_0 = 2$, and we want $N_{t=0+ \Delta t} = N_{t=0 +1} = N_1$. We apply the update rule:
+
+$$N_1 = \frac{dN}{dt} \Delta t + N_0$$
+
+$$N_1 = \frac{dN}{dt} \times 1 + N_0$$
+
+$$N_1 = \frac{dN}{dt} + N_0$$
+What should we put in for $dN/dt$? We know $dN/dt$ for the logistic equation is
```
QuLogic: Possibly a line break needed above?

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,591 @@
...
+Use Euler's method to calculate a numerical solution for the following differential equation:
+
+$$\frac{dx}{dt} = -2.3 x$$
+
+  - Try a few different step sizes ($\Delta t$) such as $0.7$ and $1$. What happens if $\Delta t$ is too large?
+
+This is what's called **numerical instability** --- the numerical solution grows very large 
+(either positive or negative) but the true solution doesn't. In the logistic equation example above,
+this numerical instability can happen if $N$ falls below $0$: the population will keep getting
+more and more negative after that.
```
QuLogic: I'm not sure I'd call that numerical instability; that's exactly the way the equation goes and it doesn't have to do with the choice of step size.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,591 @@
...
+This is pretty hard to read, especially if your calculus was a long time ago. The important thing to know for
+our purposes is that the equations we're working with are **first-order ordinary differential equations**.
+*First-order* means that the derivatives are all first derivatives (as opposed to second derivatives, third
+derivatives, etc.). **Ordinary** means that the derivatives are full derivatives (as opposed to partial
+derivatives which are written $\partial N / \partial t$). 
+
+This means that the function we'll be using is called `ode`, which stands for 'ordinary differential equation'. 
+
+```{r, eval=FALSE}
+help(ode)
```
QuLogic: Use `?ode` here too to match @joelostblom's suggestion above.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,591 @@
...
+```{r}
+# Example: deSolve with logistic eqn
+
+# now we need to define our function a bit differently to be in the format that `ode` uses
+logistic_fn <- function(t, state, parameters) {
+  # Calculates dN/dt for the logistic equation
+  
+  # t: time point at which to evaluate derivative (doesn't actually change anything in this example)
+  # state: vector of variables (here it's just N)
+  # parameters: vector of parameters c(r, K)
```
QuLogic: Maybe call it "model parameters" as opposed to function parameters.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,591 @@
...
+  N <- state 
+  
+  r <- parameters[1] # the first element of the parameters vector is r
+  K <- parameters[2] # the second element of the parameters vector is K
+
+  #rate of change
+  dN <- r * N * (1 - N / K)
+    
+  #return rate of change
+  list(c(dN))
```
QuLogic, line 409: No `return`? Did we mention implicit returns?

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,591 @@
...
+
+parameters <- c(r = 0.5, K = 50)
+state <- c(N = 10)
+times <- seq(0, 50, by = 0.01) # the timestep dt is chosen by setting the increment with 'by'
+
+#help(ode) # look at the documentation to learn about the parameters
+result <- ode(y = state, times = times, func = logistic_fn, parms = parameters)
+```
+
+The output of `ode` is a vector that contains the times we requested and its calculated values of N.
```
QuLogic: In the `class` call below, it says `matrix`, not `vector`.

lec07-pop-models.Rmd
```diff
@@ -1,4 +1,591 @@
...
+
+First, we'll simulate a trajectory for the two species.
+
+```{r}
+predator_prey <- function(t, y, parameters) {
+  # calculates dx/dt and dy/dt for a predator-prey model
+  
+  # t: time at which to evaluate derivatives
+  # y: vector of system variables (c(X, Y))
+  # parameters: vector of parameters (c(a, b, c, d))
```
QuLogic: "model parameters" as above.

QuLogic (commented):
Soo, opening lecture 7 causes RStudio to crash... Thus I can't really say much about how the result looks.

lec07-pop-models.Rmd
mbonsma: Hmmm. Good point. Maybe I should make it return something.

joelostblom (approved):

lwjohnst86 (approved):

PR #130, opened 2017-10-02
Lecture 8

mbonsma:
I wasn't able to knit this on my home machine, so there might be errors travis will find that I didn't catch. 
Will update tomorrow when I get to school!

### DIFF
--- a/lec08-stability.Rmd
+++ b/lec08-stability.Rmd
@@ -267,7 +267,7 @@ Feel free to look up a solution online.
 
 result %>% 
   rename(Prey = X, Predator = Y) %>% 
-  gather(PredPrey, PopulationSize, -time) %>% 
+  gather(PredPrey, PopulationSize, -time) %>% # the minus means use all columns except that one
   ggplot(aes(x = time, y = PopulationSize, color = PredPrey)) +
   geom_line() +
   labs(y = "Population size")
--- a/lec08-stability.Rmd
+++ b/lec08-stability.Rmd
@@ -606,7 +606,7 @@ saveGIF({
 
       p2 <- ggplot(result_i) +
         geom_path(aes(x = result_i$X, y = result_i$Y, color = result_i$time)) +
-        guides(color = guide_legend("Time")) + 
+        theme(legend.position="none") + 
         xlim(0, 8) +
         ylim(0, 8) +
         xlab("Prey population size X") +
--- a/lec08-stability.Rmd
+++ b/lec08-stability.Rmd
@@ -1,4 +1,365 @@
 ---
 title: "Fixed points, null clines, and stability"
+author: "Madeleine Bonsma-Fisher"
 ---
+
+## Lesson preamble
+
+> ### Learning objectives
+>
+> - Review numerically solving differential equations
+> - Calculate solutions with multiple initial conditions and plot
+> - Numerically solve two-dimensional models in R
+> - Find fixed points for one and two-dimensional systems analytically
+> - Find null clines for two-dimensional systems
+> - Phase portraits of two-dimensional systems
+
+> ### Lesson outline
+> 
+> Total lesson time: 2 hours
+>
+> - 
+
+> ### Setup
+> 
+> - `install.packages('deSolve')`
+> - `install.packages('ggplot2')` (or `tidyverse`)
+> - `install.packages('dplyr')` (or `tidyverse`)
+> - `install.packages('tidyr')` (or `tidyverse`)
+
+```{r math_shortcut, echo=FALSE}
+eq_dn_dt <- "$\\frac{dN}{dt}$"
+```
+
+## Recap from last week
+
+- Drawing **phase portraits** in one dimension:
+
+    - Fixed points: values of $N$ at which `r eq_dn_dt`, the rate of change of $N$,
+    is $0$. To find fixed points, plot `r eq_dn_dt` vs. $N$ and find the place(s)
+    where it crosses the $x$ axis.
+    - Stability: if you start at some $N$ close to the fixed point but not exactly
+    on it, will you go towards (stable) or away (unstable) from the fixed point? The
+    sign of `r eq_dn_dt` on either side of a fixed point tells you whether $N$ will
+    increase or decrease in that area. Draw an arrow to the right if `r eq_dn_dt` is
+    positive, and draw an arrow to the left if `r eq_dn_dt` is negative.
+
+- Numerically solving differential equations in R: starting from an initial population size, 
+calculate a sequence of population sizes using the information contained in the differential 
+equation. The result is a **trajectory** of $N$ vs. time. 
+
+    - **Euler's method**: starting from $N_0$, find the rate of change of $N$ given by `r eq_dn_dt` 
+    evaluated at $N_0$. Approximate $N_{0 + \Delta t}$ using the following expression: 
+    
+    $$N_{t+\Delta t} \approx N_t + \frac{dN}{dt} \Delta t$$
+    Use a `for` loop in R to calculate a series of $N$ at times of interest $t$. 
+    
+    - Using R's ODE-solver `ode`: define a function that calculates `r eq_dn_dt` for your model, 
+    making sure that it's in the correct format with arguments `t`, `state`, and `parameters`.
+    Call the function `ode` and give it the parameters `y = state, times = times, 
+    func = logistic_fn, parms = parameters`
+    
+## Finding numerical solutions with multiple initial conditions
+
+Let's make a pretty plot of numerical solutions for the logistic equation from a few 
+different starting points.
+
+```{r}
+# define function to be in the format that `ode` uses
+logistic_fn <- function(t, state, parameters) {
+  # Calculates dN/dt for the logistic equation
+  
+  # t: time point at which to evaluate derivative
+  # state: vector of variables (here it's just N)
+  # parameters: vector of model parameters c(r, K)
+  
+  N <- state 
+  
+  r <- parameters[1] # the first element of the parameters vector is r
+  K <- parameters[2] # the second element of the parameters vector is K
+
+  #rate of change
+  dN <- r * N * (1 - N / K)
+    
+  #return rate of change
+  return(list(c(dN)))
+}
+```
+
+
+```{r}
+# Run ode
+
+library(dplyr)
+library(tidyr)
+library(deSolve)
+
+# define parameters for the ode function
+parameters <- c(r = 0.5, K = 50)
+initial_conditions <- seq(10, 80, by = 10)
+times <- seq(0, 15, by = 0.01)
+
+# run ode inside a loop to calculate trajectories for several initial conditions
+results <- data_frame(initial_conditions = seq(10, 80, by = 10)) %>% # CTRL-shift-m to make pipe
+  # 'do' is a dplyr function that returns a dataframe after doing a series of computations.
+  do(data.frame( # 'data.frame' converts the output of 'ode' to a dataframe
+    ode(
+      y = initial_conditions, # each time the 'do' loop runs, a new initial condition will be used
+      times = times,
+      func = logistic_fn,
+      parms = parameters
+      )
+  ))
+
+# display the first few rows of the results
+head(results)  
+```
+
+This is exactly what we wanted. We have chosen a set of initial conditions 
+(`seq(10, 80, by = 10)`) and created a dataframe that has one column of times (`time`)
+and a column for each trajectory resulting from each initial condition (`X1`, `X2`, ...). 
+Now we can use `ggplot` to plot all the trajectories.
+
+```{r}
+# make a plot
+library(ggplot2)
+
+results %>%
+    gather(Column, Value, -time) %>%
+    ggplot(aes(x = time, y = Value, color = Column)) +
+    geom_line(aes(x = time, y = Value)) +
+    labs(y = "Population size")
+```
+
+This plot very nicely summarizes the behaviour of the logistic equation: it shows the
+path that the population size $N$ will take in time from several initial conditions. 
+Importantly, the carrying capacity $K$ which we set to be $50$ is clearly a stable 
+fixed point: all the trajectories go towards $K$. 
+
+## Qualitative analysis of 2D models
+
+Now we'll move on to look at how to analyze two-dimensional models. **Two-dimensional** 
+means that there are now two variables in the system (like $x$ and $y$, prey and predators) 
+instead of one (like $N$).
+
+Many of the concepts and tools from analyzing one-dimensional models will be used 
+here as well, but we will add some things that are specific to more than one dimension. 
+
+To motivate some of these new concepts and the idea of modelling in two dimensions, 
+let's analyze a model of two species interacting.
+
+$$\frac{dx}{dt} = ax - bxy$$
+$$\frac{dy}{dt} = cx - dy$$
+
+This is very close to the Lotka-Volterra predator-prey model, but the term for predator 
+growth is $cx$ instead of $cxy$. In this model the predators will eat the prey at a 
+constant rate, no matter how many of the predators there are. Make sure you don't use 
+these exact equations when you do the assignment!
+
+First, we'll simulate a trajectory for the two species.
+
+```{r}
+predator_prey <- function(t, y, parameters) {
+  # calculates dx/dt and dy/dt for a predator-prey model
+  
+  # t: time at which to evaluate derivatives
+  # y: vector of system variables (c(X, Y))
+  # parameters: vector of model parameters (c(a, b, c, d))
+  
+  # the arguments need to be named "t", "y", and "parameters", otherwise this won't work with phaseR, a package we will use later
+    
+  #now the state vector y has two elements because we have two species
+  X <- y[1] # prey
+  Y <- y[2] # predators
+  
+  a <- parameters[1]
+  b <- parameters[2]
+  c <- parameters[3]
+  d <- parameters[4]
+    
+  # calculate rate of change
+  dx <- a * X - b * Y * X
+  dy <- c * X - d * Y
+  
+  # return rate of change
+  return(list(c(dx, dy)))
+}
+```
+
+```{r}
+# run the numerical solution
+
+parameters = c(a = 5, b = 1, c = 1, d = 0.2) 
+state <- c(X = 2, Y = 2)
+times <- seq(0, 50, by = 0.01)
+
+result <- ode(y = state, times = times, func = predator_prey, parms = parameters)
+result <- data.frame(result)
+
+```
+
+```{r}
+# plot the results
+
+ggplot(result, colours()) +
+  geom_line(aes(x = time, y = X), color = 'red') +
+  geom_line(aes(x = time, y = Y), color = 'blue') +
+  labs(y = "Population size")
+```
+
+It looks like we have oscillations that decay to a fixed point. 
+
+#### Challenge
+
+Modify the plotting code above to include a legend distinguishing 'predators' and 'prey'. 
+Feel free to look up a so
+
+```{r, eval=FALSE, echo=FALSE}
+# I actually have no idea how to do this halp
+
+```
+
+*Note*: if you are using Euler's method to solve a two-dimensional system, make sure 
+you calculate the rates of change using the variables from the previous time step. For 
+example, in the predator-prey model, both $\Delta x$ and $\Delta y$ depend on $x$ and $y$, 
+and it's important that you calculate $\Delta x$ and $\Delta y$ *before* updating the 
+values of $x$ and $y$.
+
+Here's an example. 
+
+```{r}
+# Define predator-prey equations
+
+dX_dt <- function(X, Y, a, b) {
+  # X: number of prey
+  # Y: number of predators
+  # a: growth rate of prey
+  # b: rate at which predators eat prey
+  
+  return(a * X - b * Y * X)
+}
+
+dY_dt <- function(X, Y, c, d) {
+  # X: number of prey
+  # Y: number of predators
+  # c: growth rate of predators from eating prey
+  # d: death rate for predators
+  
+  return (c * X - d * Y)
+}
+```
+
+```{r}
+# parameters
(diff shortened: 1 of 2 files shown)


lec08-stability.Rmd
```diff
@@ -1,4 +1,576 @@
...
+We can also add other information, and in the plot above we haven't yet drawn arrows or
+fixed points, but we have drawn a **trajectory** - the dots trace out what happens to $X$
+and $Y$ over time starting from $X = 2$, $Y = 2$. Time is no longer an axis in the plot,
+but it's still there, hidden in the path that the populations take. 
+
+Comparing these two ways of plotting a trajectory, we can see that oscillations in time
+look like circular motion in the phase portrait. The GIF below illustrates this.
+
+```{r}
+library(animation)
```
lwjohnst86: I'm not sure why there is an error on your local machine, but the Travis error is because you are using this package without including it in the `DESCRIPTION` under `Imports:` (which you can easily add from the console via `devtools::use_package("animation")`)

lec08-stability.Rmd
```diff
@@ -1,4 +1,576 @@
...
+}
+```
+
+### Using `phaseR` to draw phase portraits in R
+
+`phaseR` is a package specifically for drawing phase portraits. It can automatically calculate 
+and plot things like trajectories and vector fields. 
+
+```{r}
+library(phaseR)
```
lwjohnst86, line 473: You'll also need to do `devtools::use_package("phaseR")` for this package as well

lec08-stability.Rmd
mbonsma: On my machine it's because ImageMagick isn't working. The space in the folder name `Program Files` is f***ing it up, and I'm not too invested in fixing it. 

I might change that block to not evaluate anyway because I usually upload the GIF separately after it's created.

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+> - Numerical solutions with multiple initial conditions (15 min)
+> - Qualitative analysis of two-dimensional models
+>      - Numerical solutions for two-dimensional models (15 min)
+>      - Phase portraits of two-dimensional systems by hand (20 min)
+>      - Using `phaseR` to create phase portraits in R (40 min)
+> - Finding fixed points analytically in one dimension (20 min)
+
+> ### Setup
+> 
+> - `install.packages('deSolve')`
```
lwjohnst86: Also `phaseR`

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+> - `install.packages('ggplot2')` (or `tidyverse`)
+> - `install.packages('dplyr')` (or `tidyverse`)
+> - `install.packages('tidyr')` (or `tidyverse`)
+
+```{r math_shortcut, echo=FALSE}
+eq_dn_dt <- "$\\frac{dN}{dt}$"
```
lwjohnst86, line 37: Handy ain't it?

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+  # Calculates dN/dt for the logistic equation
+  
+  # t: time point at which to evaluate derivative
+  # state: vector of variables (here it's just N)
+  # parameters: vector of model parameters c(r, K)
+  
+  N <- state 
+  
+  r <- parameters[1] # the first element of the parameters vector is r
+  K <- parameters[2] # the second element of the parameters vector is K
```
lwjohnst86: I thing I realized is that if `r` is *not* the first element supplied, the equation will be off. It might be better to do `paramaters['r']` to be explicit.

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+fixed point: all the trajectories go towards $K$. 
+
+#### Challenge
+
+Modify the plotting code above so that the legend lists the initial conditions for each
+line instead of `X1`, `X2`, etc. 
+
+```{r, eval=FALSE, echo=FALSE}
+# I don't know how to do this
+
```
lwjohnst86: The easiest, though not the most programmatically rigorous, is to add `mutate(Column = gsub("X", "", Column) %>% gsub("$", "0", .))`. How I see they may do it is:

```
results %>%
  rename(Condition10 = X10, Condition20 = X20 ...) %>%
  gather ...
```

Or:

```
results %>%
  mutate(Conditions = initial_conditions) %>%
  gather(Column, Value, -time, -Conditions) %>%
  ggplot(... color = Conditions) + ...
```

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+  Y <- y[2] # predators
+  
+  a <- parameters[1]
+  b <- parameters[2]
+  c <- parameters[3]
+  d <- parameters[4]
+    
+  # calculate rate of change
+  dx <- a * X - b * Y * X
+  dy <- c * X - d * Y
```
lwjohnst86: I [read](https://rstudio-pubs-static.s3.amazonaws.com/32888_197d1a1896534397b67fb04e0d4899ae.html) an easier way to do this:

```
with(as.list(c(y, parameters)), {
  dx <- a * X - b * Y * X
  dy <- c * X - d * Y
})
```
Rather than do the `a <- parameters ... b <- parameters ...`

Plus it's more resistant to different positions that the parameters are supplied.

lwjohnst86: This makes the contents of y and parameters to be accessible within the `{}`.

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+
+It looks like we have oscillations that decay to a fixed point. 
+
+#### Challenge
+
+Modify the plotting code above to include a legend distinguishing 'predators' and 'prey'. 
+Feel free to look up a solution online.
+
+```{r, eval=FALSE, echo=FALSE}
+# I actually have no idea how to do this halp
```
lwjohnst86:
```
result %>% 
  rename(Prey = X, Predator = Y) %>% 
  gather(PredPrey, PopulationSize, -time) %>% 
  ggplot(aes(x = time, y = PopulationSize, color = PredPrey)) +
  geom_line() +
  labs(y = "Population size")
```

lwjohnst86: That is the easiest way of doing it without getting into the belly of ggplot2. But this is still the best way of doing it anyway.

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+
+```
+
```
lwjohnst86: Should have an new header here.

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+# The axis was N in the 1D version. 
+# For a 2D system with variables x and y, the two axes would be x and y.
+```
+
+Continuing with the predator-prey example: instead of plotting our numerical trajectory as $X$ and 
+$Y$ vs. $t$, we'll plot $Y$ vs. $X$: predator population size vs. prey population size. 
+
+```{r}
+ggplot(result) +
+  geom_point(aes(x = X, y = Y)) # geom_line connects the dots strangely
```
lwjohnst86: Is it supposed to look strange?

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+but it's still there, hidden in the path that the populations take. 
+
+Comparing these two ways of plotting a trajectory, we can see that oscillations in time
+look like circular motion in the phase portrait. The GIF below illustrates this.
+
+![Predator-prey simulation](https://github.com/UofTCoders/rcourse/raw/master/predator-prey.gif)
+
+### Drawing qualitative phase portraits by hand
+
+Before we look at how to add more to phase portraits in R, let's pratice the process by hand to
```
lwjohnst86: s/pratice/practice

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+get familiar with it. Let's use the following two-dimensional system as an example. The two state
+variables are $x$ (position) and $v$ (velocity), and this is a model of a mass attached to a 
+spring. $dx/dt$ is the rate of change of position, and $dv/dt$ is the rate of change of velocity. 
+
+$$\frac{dx}{dt} = v$$
+
+$$\frac{dv}{dt} = -\omega^2 x$$
+
+We want to draw a **vector field** in two dimensions to show how the system will evolve in time.
```
lwjohnst86: s/evolve in time/evolve over time/ ?

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+  
+  # t: time at which to evaluate derivatives
+  # y: vector of system variables (c(x, v))
+  # parameters: vector of model parameters (c(omega))
+    
+  # the state vector y has two elements because we have two variables
+  x <- y[1] # position
+  v <- y[2] # velocity
+  
+  omega <- parameters[1]
```
lwjohnst86, line 456: As with above, you can use `with(as.list(c(y, parameters)), { .. code ..})`

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+```
+
+```{r, eval=FALSE}
+?flowField # the nice thing about this packages is that it takes functions in the same format as `ode`
+```
+
+Let's plot a vector field and some trajectories for the predator-prey model.
+
+```{r}
+library(phaseR)
```
lwjohnst86: Already loaded it above

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+                          add = FALSE)
+
+# add trajectories
+pp_trajectory <- trajectory(predator_prey, 
+                            # y0 is a matrix where each row is pairs of (X, Y) 
+                            y0 = matrix(c(1, 1, 8, 5, 2, 0, 0, 8), ncol = 2,
+                                   byrow = TRUE), 
+                            t.end = 20, # how far in time to calculate the trajectories
+                            parameters = c(a = 5, b = 1, c = 1, d = 0.2), system = "two.dim")
+```
```
lwjohnst86, line 501: oooo, these are cool plots!

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+        geom_point(aes(x = result_i$X, y = result_i$Y)) +
+        xlim(0, 8) +
+        ylim(0, 8) +
+        xlab("Prey population size X") +
+        ylab("Predator population size Y")
+    
+    grid.arrange(p1, p2, ncol = 2, top = "Predator-prey model")
+    }
+
+}, movie.name = "predator-prey.gif", ani.width = 900, ani.height = 600)
```
lwjohnst86, line 614: such a cool gif!!

lwjohnst86 (changes requested):

lec08-stability.Rmd
mbonsma: I started with that syntax, but I thought it was confusing, and to be clearer I did it the way I have it. I didn't understand the `with(as.list)` way. 

Do you have to name the parameters and stuff when you define them? i.e. does this work:

```
parameters <- (0.1, 0.5, 2, 0.1)

with(as.list(c(y, parameters)), {
dx <- a* X - b * Y * X
```

lec08-stability.Rmd
mbonsma: Uh no?

lec08-stability.Rmd
lwjohnst86: No, that way won't work. The `with` way takes the named contents of the first argument and makes them global variables in the `{}` section. You can do the same thing with a data.frame:

```
with(iris, mean(Sepal.Length))
# vs
mean(iris$Sepal.Length)
```

Yea, no worries. I understand why you are teaching this way and it makes sense. Otherwise you would have to teach about `with` and what it's doing.

lec08-stability.Rmd
lwjohnst86: Ah, right. Line assumes a continuous link for lines for each increase in x. What you want is `geom_path` which links based on the x and y pairing.

lwjohnst86: You could always change the gif to use path as well.

joelostblom:
I just tried knitting and it works fine for me (with the somewhat large caveat that `phaseR` is not available in conda so I commented out those code blocks...). Reviewing now.

lec08-stability.Rmd
joelostblom: Oh nice, they have something like that in extensions to Pandoc, but this is a neat way of doing it also

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+eq_dn_dt <- "$\\frac{dN}{dt}$"
+```
+
+## Recap from last week
+
+- Drawing **phase portraits** in one dimension:
+
+    - Fixed points: values of $N$ at which `r eq_dn_dt`, the rate of change of $N$,
+    is $0$. To find fixed points, plot `r eq_dn_dt` vs. $N$ and find the place(s)
+    where it crosses the $x$ axis.
```
joelostblom: Maybe add `(y = 0)` at the end of the sentence for clarity.

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+(`seq(10, 80, by = 10)`) and created a dataframe that has one column of times (`time`)
+and a column for each trajectory resulting from each initial condition (`X1`, `X2`, ...). 
+Now we can use `ggplot` to plot all the trajectories.
+
+```{r}
+# make a plot
+library(ggplot2)
+
+results %>%
+    gather(Column, Value, -time) %>%
```
joelostblom, line 135: Make sure to remind them that the `-` refers to all columns except the indicated one. I only mentioned that briefly.

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+    ggplot(aes(x = time, y = Value, color = Column)) +
+    geom_line(aes(x = time, y = Value)) +
+    labs(y = "Population size")
+```
```
joelostblom, line 139: Pretty!

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+
+```{r}
+# make a plot
+library(ggplot2)
+
+results %>%
+    gather(Column, Value, -time) %>%
+    ggplot(aes(x = time, y = Value, color = Column)) +
+    geom_line(aes(x = time, y = Value)) +
+    labs(y = "Population size")
```
joelostblom, line 138: I would indent the two lines under ggplot, but I know @lwjohnst86 disagrees =)

lec08-stability.Rmd
joelostblom: I think the `rename` solution is the most straightforward. For something less tedious, you could just add this

```r
names(results) <- c('time', data.matrix(results[1, -1]))
```

and then plot the same way you did earlier.

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+library(deSolve)
+
+# define parameters for the ode function
+parameters <- c(r = 0.5, K = 50)
+initial_conditions <- seq(10, 80, by = 10)
+times <- seq(0, 15, by = 0.01)
+
+# run ode inside a loop to calculate trajectories for several initial conditions
+results <- data_frame(initial_conditions = seq(10, 80, by = 10)) %>% # CTRL-shift-m to make pipe
+  # 'do' is a dplyr function that returns a dataframe after doing a series of computations.
```
joelostblom, line 110: Stress the importance and usefulness of this. Once they get their DE models into data frame format, they can now apply everything they have learned about dplyr and ggplot!

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+```{r, eval=FALSE, echo=FALSE}
+# I don't know how to do this
+
+```
+
+## Qualitative analysis of 2D models
+
+Now we'll move on to look at how to analyze two-dimensional models. **Two-dimensional** 
+means that there are now two variables in the system (like $x$ and $y$, prey and predators) 
+instead of one (like $N$).
```
joelostblom, line 182: Why is capital `N` used when there is one variable and lower case `x` and `y` with two?

lec08-stability.Rmd
mbonsma: ooooohhhhhh
!!!!

lec08-stability.Rmd
joelostblom: Is this different from the example above or could you use the same dplyr `do` approach here also?

lec08-stability.Rmd
```diff
@@ -1,4 +1,589 @@
...
+```{r}
+# plot the results
+
+ggplot(result) +
+  geom_line(aes(x = time, y = X), color = 'red') +
+  geom_line(aes(x = time, y = Y), color = 'blue') +
+  labs(y = "Population size")
+```
+
+It looks like we have oscillations that decay to a fixed point.
```
joelostblom: Can you anchor it in a real world interpretation (like the ones from last class), clarifying what it means to have partial (0.5, 1.7, etc) prey/predators?

lec08-stability.Rmd
joelostblom: She uses `geom_point` not `geom_line`, so it should be fine? It doesn't look strange to me (I think), see below.

image

You could color by time to clarify that there another variable changing:

```r
ggplot(result) +
  geom_point(aes(x = X, y = Y, color = time)) # geom_line connects the dots strangely
```

image

joelostblom: Oh just understood that your comment about `geom_line` was referring to the comment in the code. I like the dots though!

joelostblom (commented):
Looking good!!

joelostblom:
@mbonsma A few of my comments got outdated as I submitted. Gonna be faster next time, please have a look!

lec08-stability.Rmd
```diff
@@ -1,4 +1,620 @@
...
+results %>%
+    rename(Condition10 = X1, Condition20 = X2) %>% # etc
+    gather(Column, Value, -time) %>%
+    ggplot(aes(x = time, y = Value, color = Column)) +
+    geom_line(aes(x = time, y = Value)) +
+    labs(y = "Population size")
+
+# or add this line before the plot
+
+names(results) <- c('time', data.matrix(results[1, -1]))
```
joelostblom, line 173: Note that this modifies the columns of the `results` data frame, while the other two solutions only touches the modified transient data frame that is being plotted. It doesn't matter in this lecture since you don't use `results` again, but otherwise it could be worse/better depending on if you want those columns to have new names or not =p

lec08-stability.Rmd
lwjohnst86: Yes :) It is within `Ctrl-Shift-A` to format this way so I prefer this way. But up to the author! :)

lec08-stability.Rmd
lwjohnst86: Does this actually work? It's a bit unclear based on the functions. Never heard of or used `data.matrix` before.

This is a better way of doing it.

```
names(results) <- c("time", "Prey", "Predator")
# Or
results <- setNames(results, c("time", "Prey", "Predator"))
```

lec08-stability.Rmd
```diff
@@ -1,4 +1,620 @@
...
+qplot(x_seq, dx_dt) +
+  geom_hline(aes(yintercept = 0))
+```
+
+
+## Extras
+
+Predator-prey simulation GIF in R
+
+```{r}
```
lwjohnst86: I think you added this back in..? You'll need to set `eval=FALSE` to stop the CI errors.

lwjohnst86 (changes requested):

lec08-stability.Rmd
joelostblom: I just use `data.matrix` to get rid of the column names and only return the values of the first row (which are the initial conditions). You could also use `unlist(row.names=FALSE`, which is more explicit, but also more typing =) So I set the column names to '10', '20', '30', etc., not 'Prey' and 'Predator'.

joelostblom: Try it, it works!

lwjohnst86: Oh, deerrrrr, I was thinking it was a different code set... I should have looked closer.

lec08-stability.Rmd
mbonsma: Yeah this is different I think.

lec08-stability.Rmd
mbonsma: I LOVE the colouring by time!!!!

lec08-stability.Rmd
mbonsma: No reason in particular, the variables can be named anything. They could be `N1` and `N2` instead, or `predators` and `prey`, or whatever.

lec08-stability.Rmd
```diff
@@ -1,4 +1,629 @@
...
+      p1 <- ggplot(result_i) +
+        geom_line(aes(x = result_i$time, y = result_i$X), color = 'red') +
+        geom_line(aes(x = result_i$time, y = result_i$Y), color = 'blue') +
+        ylim(0, 8) +
+        xlab("Time") +
+        ylab("Population size")
+
+      p2 <- ggplot(result_i) +
+        geom_path(aes(x = result_i$X, y = result_i$Y, color = result_i$time)) +
+        guides(color = guide_legend("Time")) +
```
mbonsma: Is there a way to fix the scale for the colour legend so that it doesn't keep changing in the GIF? I couldn't figure that out.

mbonsma: Tagging @lwjohnst86

lwjohnst86: I'll look at it. Also, you don't need to use `result_i$X` inside ggplot. `X` will do.

lwjohnst86: I looked into a bit, and I know there's a way to fix that issue, but I don't readily know how without doing some digging ... and if I had time, I would dig more, but alas I don't have a lot of time :(

A easy solution would be to remove the legend entirely.

lwjohnst86: Remove legend: `theme(legend.position="none")`

lwjohnst86 (approved):

PR #131, opened 2017-10-02
Remove parts of lecture 7 not covered in class (moved to lec8)

mbonsma:
Looks like ~500 lines is about how much I get done in a lecture :)


lwjohnst86 (approved):

joelostblom:
Haha 500 is about what I settled on as well!

PR #133, opened 2017-10-03
Lecture 9

mbonsma:
Not quite finished, but almost!

### DIFF
--- a/lec09-modeling-r.Rmd
+++ b/lec09-modeling-r.Rmd
@@ -87,10 +87,10 @@ predator_prey <- function(t, y, parameters) {
   X <- y[1] # prey
   Y <- y[2] # predators
   
-  a <- parameters[1]
-  b <- parameters[2]
-  c <- parameters[3]
-  d <- parameters[4]
+  a <- parameters['a']
+  b <- parameters['b']
+  c <- parameters['c']
+  d <- parameters['d']
     
   # calculate rate of change
   dx <- a * X - b * Y * X
@@ -111,7 +111,7 @@ b <- 1
 c <- 1
 d <- 0.2
 
-params <- c(a, b, c, d)
+params <- c(a = a, b = b, c = c, d = d) # so that they can be called by name
 
 initial_conditions <-  matrix(c(1, 1, 8, 5, 2, 0, 0, 8), ncol = 2,
                                    byrow = TRUE)
@@ -154,7 +154,7 @@ back to this later when we talk about analyzing stability analytically.
 ## Null clines in two-dimensional systems
 
 A null cline is a curve on which one of derivatives in the model is $0$. In other words, 
-it's a string of points for which one of the variables is not changing --- its rate of change
+it's a series of points for which one of the variables is not changing --- its rate of change
 is $0$ on that line. 
 
 Finding null clines is similar to finding fixed points, except that when we found fixed points,
@@ -198,22 +198,24 @@ y_seq <- seq(0, 8, by = 0.1)
 df <- data.frame(x_seq, y_seq)
 
 p1 <- ggplot() + # assign plot to a variable so we can use it later
-  geom_point(aes(x = X, y = Y), result) + # trajectory
+  geom_path(aes(x = X, y = Y), result) + # trajectory
   geom_line(aes(x = 0, y = y_seq), df, color = 'red') + # x null cline 1
   geom_line(aes(x = x_seq, y = a/b), df, color = 'red') # x null cline 2
 
 p1
 ```
 
-The null clines go through the places where the trajectory 'turns back' -- they go through
+The null clines go through the places where the trajectory turns back on itself -- they go through
 the 'peaks' and 'valleys' of $x$. 
 
-Another way to think of the null clines is as if we were back to a 1D system and were
-plotting $dx/dt$ vs. $x$ for a series of values of $y$.
+Another way to think of the null clines is as if we were back to a 1D system and were 
+plotting $dx/dt$ vs. $x$ for a series of values of $y$. The null cline passes through all
+the points where $dx/dt = 0$. 
 
-Now let's find the null clines for $y$ by solving $0 = cx - dy$. In general, we want an equation in the form $y(x)$, $y$ as a function of $x$, so that we can more easily plot it on our phase portrait axes. 
+Now let's find the null clines for $y$ by solving $0 = cx - dy$. In general, we want an equation in the
+form $y(x)$, $y$ as a function of $x$, so that we can more easily plot it on our phase portrait axes. 
 
-After rearraning, we find the solution to this equation: $y = \frac{c}{d}x$. 
+After rearraning, we find the solution to this equation: $y = \frac{c}{d}x$. Let's add it to the plot
 
 ```{r}
 p1 +
@@ -224,13 +226,24 @@ p1 +
 
 The intersection(s) of the $x$ and $y$ null clines are the fixed points, so plotting null
 clines is also a way to graphically find the fixed points. This is particularly useful if 
-it's hard to solve for the fixed points. 
+it's hard to solve for the fixed points analytically. 
 
-*Note*: is there a minimum number of null clines a system must have? Are the null clines guarranteed to intersect? Are null clines always straight lines? (No, no, and no.)  
+#### Challenge (to think about)
 
-$dx/dt = 3$.  
+Is there a minimum number of null clines a system must have? Are the null clines
+guarranteed to intersect? Are null clines always straight lines?  
 
+```{r, eval=FALSE, echo=FALSE}
 
+# Consider $dx/dt = 3$. Maybe this isn't truly a differential equation, but $dx/dt$ can never be 0.
+
+# If each differential equation depends on at least one of the system variables 
+# (and the equations are different), there will always be at minimum two null clines. 
+
+# The null clines are not guarranteed to intersect: it's possible to not have a fixed point. 
+# Example: dx/dt = y, dy/dt = y + 2
+
+```
 
 #### Challenge
 
@@ -240,13 +253,113 @@ system. Verify your work by plotting the null clines.
 
 ## Quantitative stability analysis
 
-- intuition for this: assume system is linear close to the fixed point, so solutions are exponential - eigenvectors are rates of decay or growth for each eigenvector direction.
+We've now found fixed points analytically, but up until now we've been analyzing the stability 
+of the fixed points by constructing a phase portrait and looking at it. This is a perfectly good
+way to analyze stability, but we can do more: we can find out a whole lot with the quantitative
+stability analysis we're about to do:
+    - Whether a fixed point will be stable or unstable
+    - How the system will behave near a fixed point (are there oscillations, etc.)
+    - How the stability and type of a fixed point depend on the model parameters
+    
+The last reason in the list is in many ways the best reason. As we will see, in some models the
+number and/or stability of the fixed points depends on the system parameters, and quantitatively
+analyzing stability will show us this.
+
+To analyze stability, we **linearize** our model: we assume the system is linear close to the fixed
+point so that we have differential equations in the form $d\vec{x}/dt \propto \vec{x}$, where 
+$\vec{x}$ is a vector of length $D$, the number of dimensions in the model. Each element of $\vec{x}$
+is a particular direction in phase space. In one dimension it's the same as the model variable, 
+but in more than one dimension the directions don't necessarily have to be the same as the model
+variables. Because the equations are in this linearized form, their solutions are exponential 
+functions. These solutions will tell us how the system behaves in the region close to the fixed 
+points. 
+
+That was a lot. Let's start looking at examples. 
+
+### Quantitative stability analysis in one dimension
+
+Let's look at the logistic equation. 
+
+$$\frac{dN}{dt} = rN (1-\frac{N}{K})$$
+
+Let's call the right-hand side of this equation $f(N)$: $f(N) = rN(1 - \frac{N}{K})$. First, we find the fixed points by setting $f(N) = 0$. $N = 0$ and $N = K$ are the fixed points.
+
+Next, to linearize this equation, we take the **derivative** of $f(N)$ with respect to $N$: we calculate $\frac{\partial f(N)}{\partial N}$. 
+
+$$f(N) = rN(1 - \frac{N}{K}) = rN - \frac{rN^2}{K}$$.
+
+In the second form it's a bit easier to take the derivative with respect to $N$.
+
+$$\frac{\partial}{\partial N} f(N) = r - \frac{2rN}{K}$$
+
+*Note*: the symbol $\partial$ stands for 'partial derivative': it means that we only worry about terms that **explicitly** depend on $N$ in this case. 
+
+#### Challenge
+
+Evaluate the following derivatives:
+
+$$\frac{d}{dx} x^2$$
+
+$$\frac{\partial}{\partial N} \text{e}^N$$
+
+$$\frac{\partial}{\partial x} \frac{y}{x}$$
+
+Now comes the magic. We evaluate the derivative we just calculated, $\frac{\partial f(N)}{\partial N}$,
+at each of the fixed points in turn. If it's **negative**, the fixed point is **stable**. If it's
+**positive**, the fixed point is **unstable**. That's all there is to it!
+
+For the fixed point $N = 0$: 
+
+$$\frac{\partial}{\partial N} f(N) \vert_{N=0} = (r - \frac{2rN}{K})\vert_{N=0} = r$$
+
+$r$ is a number we haven't chosen, but we do know that it's always positive. That means $N = 0$ is an unstable fixed point. 
+
+For the fixed point $N = K$: 
+
+$$\frac{\partial}{\partial N} f(N) \vert_{N=K} = (r - \frac{2rN}{K})\vert_{N=K} = r - \frac{2rK}{K} = -r$$
+
+If $r$ is positive, then this expression is negative, and $N=K$ is a stable fixed point.
+Notice that if we happened to make the parameter $r$ negative, the stability of the fixed points 
+would switch! These kinds of observations are much easier to make after we've quantitatively analyzed
+stability.
+
+Here's another way to think about what just happened. When we were drawing phase portraits, we looked at
+the sign of $\frac{dN}{dt}$ and drew arrows to the left or right depending on its sign. Now, by looking
+at $\frac{\partial}{\partial N} \frac{dN}{dt}$ instead, we're formally calculating the *slope* of
+$\frac{dN}{dt}$ with respect to $N$ near the fixed point: where the slope is negative, we know that 
+$N$ will be drawn back towards the fixed point, and where the slope is positive, $N$ will be pushed 
+away from the fixed point. 
+
+Let's revisit the differential equation $\frac{dx}{dt} = \text{sin}(x)$ to explore this idea.
+
+```{r}
+x_seq <- seq(0, 10, by = 0.1)
+
+qplot(x_seq, sin(x_seq)) +
+  geom_hline(yintercept = 0)
+```
+
+The four fixed points visible here are respectively unstable, stable, unstable, and stable. 
+
+### Two dimensions: calculating the Jacobian matrix
+
+Analyzing stability in two dimensions follows exactly the same principle, but the math is a little bit more complicated. Now instead of needing to find the slope in just one dimension, we have an two-dimensional surface we need to consider. 
+
+Imagine the landscape of $dx/dt$ and $dy/dt$ as an actual landscape with hills and valleys. We could have mountain peaks that represent unstable fixed points:
+
+![Mt Rundle, Alberta](https://chrismartinphotography.files.wordpress.com/2014/09/mount-rundle-moonrise-c2a9-christopher-martin-95461.jpg)
+
+We could have beautiful valleys that represent stable fixed points:
+
+![Evergreen Brickworks](https://www1.toronto.ca/City%20Of%20Toronto/Parks%20Forestry%20&%20Recreation/Events/images/brickworks-park-headerimages.jpg)
+
+We could have a treacherous mountain pass:
+
 
 Mountain pass: https://en.wikipedia.org/wiki/Mountain_pass
 
-### Calculating the Jacobian matrix
-
 ### Classifying fixed points using eigenvalues of the Jacobian
+
 
 
 ### An example in three dimensions (bac-virus-nutrients)
--- a/lec09-modeling-r.Rmd
+++ b/lec09-modeling-r.Rmd
@@ -22,6 +22,7 @@ author: "Madeleine Bonsma-Fisher"
 
 > ### Setup
 > 
+> - `install.packages('phaseR')`
 > - `install.packages('deSolve')`
 > - `install.packages('ggplot2')` (or `tidyverse`)
 > - `install.packages('dplyr')` (or `tidyverse`)
@@ -104,6 +105,7 @@ predator_prey <- function(t, y, parameters) {
 Now we create the phase portrait using `phaseR`, add some trajectories, and add the fixed points.
 
 ```{r}
+library(phaseR)
 
 # parameters
 a <- 5
@@ -178,6 +180,7 @@ clines we just found.
 
 ```{r}
 # run the numerical solution
+library(deSolve)
 
 parameters = c(a = 5, b = 1, c = 1, d = 0.2) 
 state <- c(X = 2, Y = 2)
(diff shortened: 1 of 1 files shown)


lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+the 'peaks' and 'valleys' of $x$. 
+
+Another way to think of the null clines is as if we were back to a 1D system and were 
+plotting $dx/dt$ vs. $x$ for a series of values of $y$. The null cline passes through all
+the points where $dx/dt = 0$. 
+
+Now let's find the null clines for $y$ by solving $0 = cx - dy$. In general, we want an equation in the
+form $y(x)$, $y$ as a function of $x$, so that we can more easily plot it on our phase portrait axes. 
+
+After rearraning, we find the solution to this equation: $y = \frac{c}{d}x$. Let's add it to the plot
```
QuLogic: s/rearraning/rearranging/

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+           # y0 is a matrix where each row is pairs of (X, Y) 
+           y0 = initial_conditions, 
+           t.end = 30, # how far in time to calculate the trajectories 
+           parameters = params, system = "two.dim")
+```
+
+#### Challenge (to think about)
+
+Is there a minimum number of null clines a system must have? Are the null clines
+guarranteed to intersect? Are null clines always straight lines?
```
QuLogic: s/guarranteed/guaranteed/

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+
+$$0 = x(a-by)$$
+
+$$0 = cx - dy$$
+
+One possibility from the first equation is $x = 0$. If $x = 0$, then $y$ must also be $0$
+in order for $cx - dy = -dy = 0$. So one fixed point is $x = 0, y = 0$. 
+
+The other factor in the first equation gives $a = by$, so $y = a/b$. If we plug $y = a/b$
+into the second equation, we get $cx = da/b$, so $x = \frac{da}{bc}$. The second fixed point
```
QuLogic: I'd switch to `ad` instead of `da` to avoid confusion with the differential since we don't really change font styles. It also matches the code later.

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+# Challenge solution - basically copy and paste from lecture 8
+
+predator_prey <- function(t, y, parameters) {
+  # calculates dx/dt and dy/dt for a predator-prey model
+  
+  # t: time at which to evaluate derivatives
+  # y: vector of system variables (c(X, Y))
+  # parameters: vector of model parameters (c(a, b, c, d))
+  
+  # the arguments need to be named "t", "y", and "parameters", otherwise this won't work with phaseR, a package we will use later
```
QuLogic, line 132: That one problem that someone had during the challenge today was basically due to this thing.

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+
+# x = 0, y = 0 
+points(0, 0, # x, y coordinates of point to plot
+       cex = 1.5) # 'cex' stands for 'character expansion': marker size relative to default
+
+# x = ad/bc, y = a/b 
+x_star <- a * d / (b * c)
+y_star <- a / b
+points(x_star, y_star, # x, y coordinates of point to plot
+       pch = 21, # pch = 21 gives a filled circle
```
QuLogic: Maybe mention that `pch` means 'plotting character'.

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+       bg = 'black', # background colour
+       cex = 1.5) 
+```
+
+One interesting thing about this system is that $x = 0, y = 0$ is an unstable fixed point, but
+there is one direction in which trajectories are attracted to it: along the y-axis. This makes
+sense: if there are no prey ($x = 0$), then the predators must also go extinct. We will come
+back to this later when we talk about analyzing stability analytically.
+
+## Null clines in two-dimensional systems
```
QuLogic: Wikipedia seems to think nullcline is one word...

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+
+A null cline is a curve on which one of derivatives in the model is $0$. In other words,
```
QuLogic: s/one of/one of the/

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+guarranteed to intersect? Are null clines always straight lines?  
+
+```{r, eval=FALSE, echo=FALSE}
+
+# Consider $dx/dt = 3$. Maybe this isn't truly a differential equation, but $dx/dt$ can never be 0.
+
+# If each differential equation depends on at least one of the system variables 
+# (and the equations are different), there will always be at minimum two null clines. 
+
+# The null clines are not guarranteed to intersect: it's possible to not have a fixed point.
```
QuLogic: s/guarranteed/guaranteed/

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+
+# The null clines are not guarranteed to intersect: it's possible to not have a fixed point. 
+# Example: dx/dt = y, dy/dt = y + 2
+
+```
+
+#### Challenge
+
+Find the fixed point(s) and null clines analytically for the following two-dimensional 
+system. Verify your work by plotting the null clines.
```
QuLogic: I guess you haven't added it yet?

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+qplot(x_seq, sin(x_seq)) +
+  geom_hline(yintercept = 0)
+```
+
+The four fixed points visible here are respectively unstable, stable, unstable, and stable. 
+
+### Stability analysis in two dimensions
+
+Analyzing stability in two dimensions follows exactly the same principle, but the math is a little bit 
+more complicated. Now instead of needing to find the slope in just one dimension, we have an
```
QuLogic: s/an/a/

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+# Challenge solution
+
+# a fixed point like a mountain pass is called a 'saddle point', and it's unstable. 
+# It has one stable direction (from the peak down into the pass) and one unstable direction (down from the pass)
+```
+
+### Calculating the Jacobian matrix
+
+We will now learn how to formally describe the stability of fixed points in two-dimensional systems --- 
+how to quantitativley distinguish all the possible landscape shapes that can surround a fixed point.
```
QuLogic: s/quatitativley/quantitatively/

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+
+This is the Jacobian matrix. It's often called $J$.
+
+$$J=\begin{bmatrix}
+    \frac{\partial f}{\partial x} & \frac{\partial f}{\partial y} \\
+    \frac{\partial g}{\partial x} & \frac{\partial g}{\partial y}
+\end{bmatrix}$$
+
+*Note*: it doesn't matter which variable is which, or which equation is called $f$ or $g$, as 
+long as the order consistent --- one function on the top row, the other on the bottom, one
```
QuLogic: s/order/order is/

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+
+Our challenge is to find the eigenvalues of the Jacobian. We could also find the corresponding
+eigenvectors, but we don't need to for stability analysis.
+
+There is a formula to calculate eigenvalues: solve the equation $\text{Det}(A - \lambda \mathbb{1}) = 0$).
+$\text{Det}$ stands for 'determinant', and $\mathbb{1}$ is the **identity matrix**: a matrix of the same
+shape as $A$ but with $1$s on the diagonal and $0$s everywhere else.
+
+However, we won't calculate eigenvalues by hand. R has a function called `eigen` that we can use to
+directly calculate the eigenvalues of a matrix.
```
QuLogic, line 596: Mention that it's for one fixed point at a time first ...

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,567 @@
...
+
+There are two eigenvalues because our system is two-dimensional. What do they mean? For a set of eigenvalues ($\lambda_1$, $\lambda_2$, etc.):
+
+ - If the real part of $\lambda < 0$, that eigendirection is **stable**.
+ - If the real part of $\lambda > 0$, that eigendirection is **unstable**.
+ - If all eigenvalues have negative real parts, the fixed point is **stable**.
+ - If one or more eigenvalues has a positive real part, the fixed point is **unstable**.
+ - If the eigenvalues have an imaginary (**complex**) part, the fixed point is **oscillatory**.
+
+Let's do the other fixed point:
```
QuLogic, line 643: ... instead of after.

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+> - Analyze stability for one and two-dimensional systems analytically
+>     - Write down the Jacobian matrix
+>     - Use R to find eigenvalues of the Jacobian
+
+> ### Lesson outline
+> 
+> Total lesson time: 2 hours
+>
+> - Review and warm-up challenge (10 minutes)
+> -
```
joelostblom: Don't forget to complete this

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+
+That was a lot. Let's start looking at examples. 
+
+### Quantitative stability analysis in one dimension
+
+Let's look at the logistic equation. 
+
+$$\frac{dN}{dt} = rN (1-\frac{N}{K})$$
+
+Let's call the right-hand side of this equation $f(N)$: $f(N) = rN(1 - \frac{N}{K})$. First, we find the fixed points by setting $f(N) = 0$. $N = 0$ and $N = K$ are the fixed points.
```
joelostblom: $N = 0$ and $N = K$ are the fixed points. `->` The fixed points are $N = 0$ and $N = K$.

Just so that the last equation from the previous sentence does not blend in with this equation.

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+
+![Evergreen Brickworks](https://www1.toronto.ca/City%20Of%20Toronto/Parks%20Forestry%20&%20Recreation/Events/images/brickworks-park-headerimages.jpg)
+
+We could have a treacherous mountain pass:
+
+![Abra Malaga Pass, Peru](https://d36tnp772eyphs.cloudfront.net/blogs/1/2015/06/4a.Peru-Abra-Malaga2.jpg)
+
+Here's an idealized representation of a mountain pass. 
+
+![Idealized mountain pass, By Paulba Legend (talk), CC BY-SA 3.0](https://en.wikipedia.org/wiki/Mountain_pass#/media/File:Saddleroute3.JPG)
```
joelostblom: Does not render. Change the image address to `https://upload.wikimedia.org/wikipedia/commons/2/29/Saddleroute3.JPG`

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+dx_dx <- a - b*y
+dx_dy <- -b*x
+dy_dx <- c
+dy_dy <- -d
+
+# create the Jacobian matrix
+J<-matrix(c(dx_dx, dx_dy, dy_dx, dy_dy), nrow = 2, ncol = 2, byrow=TRUE);
+
+# calculate eigenvalues
+e_vals<-eigen(J)$values
```
joelostblom: spaces around `<-`

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+x <- 0
+
+# define the partial derivatives 
+dx_dx <- a - b*y
+dx_dy <- -b*x
+dy_dx <- c
+dy_dy <- -d
+
+# create the Jacobian matrix
+J<-matrix(c(dx_dx, dx_dy, dy_dx, dy_dy), nrow = 2, ncol = 2, byrow=TRUE);
```
joelostblom: spaces around `<-`

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+
+# calculate eigenvalues
+e_vals<-eigen(J)$values
+```
+
```
joelostblom: Maybe show them the plot again here and relate the results from the calculations above to the plot to ensure that everyone understands how the numbers translate into what we have done graphically so far.

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+    a - by & -bx \\
+    c & -d
+\end{bmatrix}$$
+
+### Classifying fixed points using eigenvalues of the Jacobian
+
+From here, we need to calculate the **eigenvalues** of the Jacobian matrix. The eigenvalues of a
+square matrix $A$ are numbers (often denoted $\lambda$) that satisfy the following relationship,
+where $\vec{v}$ is a vector whose length is the dimension of $A$. If $A$ is a 2-by-2 matrix, 
+$\vec{v}$ is a length-2 vector.
```
joelostblom: ...whose length is the dimension of $A$. If $A$ is a 2-by-2 matrix, +$\vec{v}$ is a length-2 vector.

`->`

...whose length is the dimension of $A$ (e.g. if $A$ is a 2-by-2 matrix, +$\vec{v}$ is a length-2 vector.)

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+square matrix $A$ are numbers (often denoted $\lambda$) that satisfy the following relationship,
+where $\vec{v}$ is a vector whose length is the dimension of $A$. If $A$ is a 2-by-2 matrix, 
+$\vec{v}$ is a length-2 vector. 
+
+$$A\vec{v} = \lambda \vec{v}$$.
+
+What this means is that if the matrix $A$ is multiplied with a particular vector $\vec{v}$,
+the resulting vector is a number multiplied by the original vector. 
+
+This isn't true for just any vector. In general, this is how matrix multiplication works:
```
joelostblom: It is not true for every matrix either, is it? Is it more suitable to say matrix here, since we are calculating the eigenvalue of the Jacobian matrix?

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+
+- Whether a fixed point will be stable or unstable
+- How the system will behave near a fixed point (are there oscillations, etc.)
+- How the stability and type of a fixed point depend on the model parameters
+    
+The last reason in the list is in many ways the best reason. As we will see, in some models the
+number and/or stability of the fixed points depends on the system parameters, and quantitatively
+analyzing stability will show us this.
+
+To analyze stability, we **linearize** our model: we assume the system is linear close to the fixed
```
joelostblom, line 385: Is it important to decide on *how* close to the fix point? I guess the curves here are quite smooth, but say we actually look at a mountain pass, then there would be many small bumps along the way but the overall trajectory would still be done towards the valley, are there rules for how to chose the step size to capture the overall direction? Maybe worth mentioning something about it without going into details?

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+                            parameters = params, system = "two.dim")
+
+pp_nullclines <- nullclines(predator_prey, 
+           x.lim = c(0, 10), # axis limits
+           y.lim = c(0, 10),
+           # y0 is a matrix where each row is pairs of (X, Y) 
+           y0 = initial_conditions, 
+           t.end = 30, # how far in time to calculate the trajectories 
+           parameters = params, system = "two.dim")
+```
```
joelostblom, line 314: Why is the red line not covering the entire range of `y` when `x=0`? And why is there that little dip in when `x=1` and `y=5`?

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+y_seq <- seq(0, 8, by = 0.1)
+
+df <- data.frame(x_seq, y_seq)
+
+p1 <- ggplot() + # assign plot to a variable so we can use it later
+  geom_path(aes(x = X, y = Y, color = time), result) + # trajectory
+  geom_line(aes(x = 0, y = y_seq), df, color = 'dark red') + # x null cline 1
+  geom_line(aes(x = x_seq, y = a/b), df, color = 'dark red') # x null cline 2
+
+p1
```
joelostblom: Is understand both lines analytically, but only the y=5 line from the graph... The system seem to never touch the x=0 line, is that just a coincidence based on the values we chose?

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+to our predator-prey model.
+
+$$J=\begin{bmatrix}
+    a - by & -bx \\
+    c & -d
+\end{bmatrix}$$
+
+### Classifying fixed points using eigenvalues of the Jacobian
+
+From here, we need to calculate the **eigenvalues** of the Jacobian matrix. The eigenvalues of a
```
joelostblom, line 547: Can you add a colloquial/graphical explanation of why the eigenvalues will tell us the slope/stability of the fix points?

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+dy_dy <- -d
+
+# create the Jacobian matrix
+J<-matrix(c(dx_dx, dx_dy, dy_dx, dy_dy), nrow = 2, ncol = 2, byrow=TRUE);
+
+# calculate eigenvalues
+e_vals<-eigen(J)$values
+```
+
+There are two eigenvalues because our system is two-dimensional. What do they mean? For a set of eigenvalues ($\lambda_1$, $\lambda_2$, etc.):
```
joelostblom: Is there a good explanation/image/animation that will help give intuition about this? *Why* do the eigenvalues infer these things about the fixed points?

joelostblom (changes requested):
This is good! Would be great for some more colloquial explanation to give intuition about how eigenvalues relate to fixed points (if possible)

lec09-modeling-r.Rmd
lwjohnst86: I agree, I was a bit confused by this. Same with `dy` on line 94 above.

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+that we found *all* the fixed points, even ones that we wouldn't find by simulating trajectories. 
+
+Finding fixed points in two dimensions is similar to finding them in one dimension, but now
+we have two equations that we have to solve *simultaneously*. That means we set both $dx/dt = 0$
+and $dy/dt = 0$ and find combinations of $x$ and $y$ that make both zero at the same time. 
+
+Let's find the fixed points for our predator-prey model:
+
+$$\frac{dx}{dt} = ax - bxy$$
+$$\frac{dy}{dt} = cx - dy$$
```
lwjohnst86: see @QuLogic and my note below. `dy` could be something like `By`, so it's different from `dy/dt`

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+```{r}
+library(phaseR)
+
+# parameters
+a <- 5
+b <- 1
+c <- 1
+d <- 0.2
+
+params <- c(a = a, b = b, c = c, d = d) # so that they can be called by name
```
lwjohnst86, line 163: why not do the parameter values inside the `c()`? e.g. `c(a = 5, b = 1 ...)`?

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+
+initial_conditions <-  matrix(c(1, 1, 8, 5, 2, 0, 0, 8), ncol = 2,
+                                   byrow = TRUE)
```
lwjohnst86: could this be represented a bit differently? trying imagine a matrix in flat format is difficult... e.g. is `1, 1` the first row or is `1, 2` the first row? Maybe something like:

```
matrix(
    c(1, 1,
      8, 5,
      2, 0,
      0, 8), 
    ncol = 2,
    byrow = TRUE)
```

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+pp_trajectory <- trajectory(predator_prey, 
+                            # y0 is a matrix where each row is pairs of (X, Y) 
+                            y0 = initial_conditions, 
+                            t.end = 30, # how far in time to calculate the trajectories
+                            parameters = params, system = "two.dim")
+
+# plot fixed points
+
+# x = 0, y = 0 
+points(0, 0, # x, y coordinates of point to plot
```
lwjohnst86: Maybe mention that this is using the base plotting system, which they have not learned and which we *strongly* recommend not using in their final assignment. But that because `phaseR` uses base plot that you need to here.

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+For any value of $x$, provided $y = a/b$, $x$ won't change. 
+
+Let's run a numerical solution for the predator-prey model we've been using and plot the null
+clines we just found.
+
+```{r}
+# run the numerical solution
+library(deSolve)
+
+parameters = c(a = 5, b = 1, c = 1, d = 0.2)
```
lwjohnst86: = to <-

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+
+x_seq <- seq(0, 8, by = 0.1)
+y_seq <- seq(0, 8, by = 0.1)
+
+df <- data.frame(x_seq, y_seq)
+
+p1 <- ggplot() + # assign plot to a variable so we can use it later
+  geom_path(aes(x = X, y = Y, color = time), result) + # trajectory
+  geom_line(aes(x = 0, y = y_seq), df, color = 'dark red') + # x null cline 1
+  geom_line(aes(x = x_seq, y = a/b), df, color = 'dark red') # x null cline 2
```
lwjohnst86: where did `a/b` come from? All the way at the top? Might be good to re-include it in this code chunk.

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+
+Evaluate the following derivatives:
+
+$$\frac{d}{dx} x^2$$
+
+$$\frac{\partial}{\partial N} \text{e}^N$$
+
+$$\frac{\partial}{\partial x} \frac{y}{x}$$
+
+Now comes the magic. We evaluate the derivative we just calculated, $\frac{\partial f(N)}{\partial N}$,
```
lwjohnst86: Is this part of the challenge? If not, should put a header above it.

lec09-modeling-r.Rmd
```diff
@@ -1,4 +1,615 @@
...
+x <- d * a / (b * c)
+
+# define the partial derivatives 
+dx_dx <- a - b*y
+dx_dy <- -b*x
+dy_dx <- c
+dy_dy <- -d
+
+# create the Jacobian matrix
+J<-matrix(c(dx_dx, dx_dy, dy_dx, dy_dy), nrow = 2, ncol = 2, byrow=TRUE);
```
lwjohnst86: spaces

lwjohnst86 (commented):
Looks good! Some minor comments/edits.

lec09-modeling-r.Rmd
mbonsma: The x = 0 line is sort of special in that it's a stable direction - if you scroll up two plots you can see that trajectories that start on that line stay on it and go to x = 0, y = 0.

lec09-modeling-r.Rmd
mbonsma: I also noticed that dip, I don't think it's real though because if you change the scale it doesn't change size. I don't know why it's there, or why the red line doesn't go all the way to y = 0.

lec09-modeling-r.Rmd
mbonsma: Yeah good point. In quantitative analysis you don't have to decide how close is close enough, because you're calculating the slope exactly at the fixed point.

lec09-modeling-r.Rmd
mbonsma: You can always find eigenvalues / vectors for a square matrix, but they might be degenerate or zero or something.

lec09-modeling-r.Rmd
mbonsma: Yeah you've mentioned this a few times. I'm not sure how best to show it - in my opinion the clearest explanation of why this works is quite technical. I'll see what I can do!

lec09-modeling-r.Rmd
mbonsma: Because I want to use them individually a bit further down...

lec09-modeling-r.Rmd
mbonsma: I like it!

PR #142, opened 2017-10-10
Edit lecture 17 description

linamnt:
is this the only place it would be?


lwjohnst86:
It's also in `_site.yml`, line 45. But you could change that when you create the lesson.

PR #143, opened 2017-10-10
Lecture 11

lwjohnst86:


lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
+---
+title: "Statistical/probabilistic modelling in R"
```
joelostblom: 'and' instead of '/'?

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+types of analyses. First, let's load up a dataset.
+
+```{r, message=FALSE}
+portal <- read_csv("portal_data.csv")
+```
+
+Now, let's fit a model to the data. Let's say we want to see the role that sex
+has on weight in various species in the `portal` dataset. (`gaussian` is stating 
+that `weight` is a continuous variable and that you assume the error terms have
+a gaussian, or normal, distribution.)
```
joelostblom: **G**aussian

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+fit2 <- glm(weight ~ sex + hindfoot_length, data = portal)
+tidy(fit2)
+```
+
+$$ weight = -32.09 + (sexM \times -1.72) + (hindfoot\_length \times 2.56)$$
+
+Hmm, that's interesting... is a *negative* intercept... How can weight be
+negative? Well, that's because the intercept is calculated by setting *both* sex
+and hindfoot length to zero! But is it physically even possible that a hindfoot be
+non-existant?? No. But math doesn't care what is physically possible or not.
```
joelostblom: non-exist**e**nt

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+That's where your ecology/biology knowledge comes in. So let's change that up by
+changing the actual value of hindfoot length in the dataset. There are several
+ways to transform the data into something that is meaningful to us. But first:
+
+> What does everyone think is a meaningful way to transform hindfoot length so
+we can interpret the model more appropriately?
+
+There are many ways to do it, but one way that is often used is to mean center
+the data. Mean centering is substracting each value by the mean. After mean
```
joelostblom: subtracting

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+    select(term, estimate, conf.low, conf.high, p.value)
+```
+
+In this case, the uncertainty of the estimate for hindfoot length is between
+2.524 to 2.58, which is pretty narrow (that's a good thing). This tells us that
+the estimate that hindfoot length influences weight is probably reflective of
+reality. That makes sense since longer hindfeet mean there is more space and
+amount of area for more flesh to attach to and that there is more bone, which
+adds weight. We can also be more strict about our uncertainty
+(counterintuitively, a larger confidence interval is more strict).
```
joelostblom: counter-intuitively

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+on the discoveries physicists made when they created a visualization of a black
+hole based on the math when making the movie Interstellar).
+
+But even the type of graph you use can hide data. Check out the figure below for
+an example of that. In general, it's better to visualize the raw data points,
+but this quickly becomes a problem as you get more data.
+
+![Visualizing raw data vs aggregate](boxplot-problem.gif)
+
+Another example is known as Simpon's Paradox. You all have already encountered
```
joelostblom: Simp**s**on's

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+    theme(legend.position = "none") +
+    facet_wrap( ~ dataset, ncol = 3)
+```
+
+And sometimes, by not looking at the raw data, it can lead to actual harm.
+For instance, Simpson's paradox comes up in medical studies of drugs
+that could be lifesaving for a patient. If inappropriate conclusions are drawn
+because the data wasn't completely explored, people could be harmed or could
+die. This is true in *any* scientific field, including ecology. Policies aimed
+at benefitting the environment could in fact be *harming* it because these types
```
joelostblom: benefiting

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+library(tidyverse)
+library(broom)
+library(datasauRus)
+```
+
+## Statistics and probability
+
+Theoretical models are powerful tools at explaining or understanding the world.
+However, they are limited in that the real-world often doesn't perfectly fit
+these models. The real world is messy and noisy. We can't always trust our data
```
joelostblom: maybe add 'blindly' `->` "we can't always blindly trust our data"

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+
+Or, a simplified, alphabetized version is:
+
+$$ y = a + Xb + e $$
+
+Where $a$ is the intercept, $X$ is the variable, $b$ is the slope/coefficient,
+and $e$ is the error term. In the case of multiple linear regression (more than
+one $x$), it is expanded to:
+
+$$ y = a + x_1b_1 + x_2b_2 +...+ x_nb_n + e $$
```
joelostblom: Why small `x` here, and capital `X` when there is only one `b`?

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+correlation or analysis of variance (ANOVA), linear regression, logistic
+regression. GLM is a powerful technique that you can use on a wide range of data
+and research questions, and is the foundation for understanding how more
+advanced techniques work. This is the reason I will be covering GLM in depth.
+(For those who will go into graduate students or into research in general, you
+will likely use mixed effects models. If you can understand and grasp GLM, it
+will be substantially easier for you to understand mixed effects modelling and
+other techniques.)
+
+A version of GLM that uses continuous $y$ values is called linear regression,
```
joelostblom, line 80: I would briefly mention that the name linear regression is not intuitive, just so that they don't try to understand why it is called that. Optionally, you could also include the etymology here or later, [according to wikipedia](https://en.wikipedia.org/wiki/Linear_regression#Estimation_methods):

> Francis Galton's 1875 illustration of the correlation between the heights of adults and their parents. The observation that adult children's heights tended to deviate less from the mean height than their parents suggested the concept of "regression toward the mean", giving regression its name.
>

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+which I'm going to focus on. The formula for linear regression (or GLM in
+general) is:
+
+$$ Y = \alpha + X\beta + \varepsilon $$
```
joelostblom, line 84: Do you think it is worthwhile to include a plot with a linear regression line somewhere before you get into the coding so that they can visualize it when you talk about intercepts and slopes later on?

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+
+We can also expand the model to include more terms.
+
+```{r}
+fit_hfsex <- glm(weight ~ hindfoot_length + sex, data = portal)
+tidy(fit_hfsex, conf.int = TRUE) %>% 
+    select(term, estimate, conf.low, conf.high, p.value)
+```
+
+Great! So in almost all cases, you should favour confidence intervals and
```
joelostblom: Remove "So"

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+you ask a question such as "Are males heavier than females in rodents?". There
+is then the null hypothesis ("females and males weigh the same") and the
+alternative hypothesis ("males are heavier than females"). Using statistical
+tests, we can calculate the p-value to give us an indication of which hypothesis
+is true. Traditionally, if the p-value is <0.05 you 'reject' the null
+hypothesis.
+
+However, the true meaning of the p-value is a bit counter intuitive. The
+p-value is the probability of getting a result that is the same as or more
+extreme as what you are getting given the null hypothesis is true. There is
```
joelostblom: During the lecture, include an example/rephrasing here, something like "If it is true that females and males weight the same, what is the probability that we would get this (or a more extreme) outcome?"

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+
+qplot(y = p.values) +
+    geom_hline(yintercept = 0.05)
+```
+
+As you can see, we *know* that x and y are the same (both have a mean of 0 and
+standard deviation of 1). And yet, there are still ~6 or so (below the line)
+tests that show a "significant" difference between x and y! Which makes sense,
+since a p-value of 0.05 equals a 5% chance you will see another similar result
+if the null hypothesis were true. Confusing eh?
```
joelostblom, line 401: Could be interesting to mention that 5% [really is just an arbitrary choice](http://www.p-value.info/2013/01/whats-significance-of-005-significance_6.html) and ask them if they think it makes sense, should it be different for different fields (bigger consequences requires smaller p-value?). Does the size of the observed effect impact which the interpretation of the results in addition to p-value?

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+```{r}
+exp(-1.035 + (-0.0003 * 50) + (-0.0338 * 0) + (-0.0004 * 30))
+```
+
+In this case, for a female with a weight of 50 and hindfoot of 30, the odds of
+disease is 0.345, fairly small. Interpretation is actually a bit more
+complicated than that. So depending on your final project, if you need/want to
+use this, we can help walk you through this more.
+
+## Confidence intervals, and p-values
```
joelostblom, line 304: I think it would be interesting to mention effect size also. Just briefly remind them that no matter their p-value or confidence interval, it is also important to assess whether the observed differences in the measurement could lead to actual biological differences in the process the study.

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+    x <- rnorm(sample_size, mean = 0, 
+               sd = variability)
+    y <- rnorm(sample_size, mean = 0 + actual_difference, 
+               sd = variability)
+    tidy(t.test(x, y))[['p.value']]
+})
+
+qplot(y = p.values) +
+    geom_hline(yintercept = 0.05)
+```
```
joelostblom, line 395: Include link to dance of the p-values if they want to read more https://www.youtube.com/watch?v=5OL1RqHrZQ8

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+model can provide. (For a real-world example of this, check out this 
+[article](https://www.wired.com/2014/10/astrophysics-interstellar-black-hole/)
+on the discoveries physicists made when they created a visualization of a black
+hole based on the math when making the movie Interstellar).
+
+But even the type of graph you use can hide data. Check out the figure below for
+an example of that. In general, it's better to visualize the raw data points,
+but this quickly becomes a problem as you get more data.
+
+![Visualizing raw data vs aggregate](boxplot-problem.gif)
```
joelostblom: The gif is hypnotizing! I could watch it for hours =) I don't completely understand how it is showing that thhe distributions differ though, do you have a link to where you got it so that I can read more?

joelostblom (commented):
yay, stats!

lec11-modelling.Rmd
lwjohnst86: There is a reason (X is a matrix of all the variables, while x is a single variable), but it isn't something that is important. I'll change it to small x.

lec11-modelling.Rmd
lwjohnst86: I've got that link in the resources below! I was planning on making a suggestion to see that link for more detail

lec11-modelling.Rmd
lwjohnst86: Oh, shiiittt, right, I forgot about including a citation to it.. oops! I'll add it.

The changes in the distributions show that as the distribution changes, the boxplots do *not* change! So boxplots can be tricky too.

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+
+```{r setup}
+library(tidyverse)
+library(broom)
+library(datasauRus)
+```
+
+## Statistics and probability
+
+Theoretical models are powerful tools at explaining or understanding the world.
```
mbonsma: at -> for
or -> our
understanding of

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+
+## Generalized linear models
+
+Generalized linear models (or GLM) is a family of statistical modelling
+techniques, that can also be used in place of several other techniques, e.g.
+correlation or analysis of variance (ANOVA), linear regression, logistic
+regression. GLM is a powerful technique that you can use on a wide range of data
+and research questions, and is the foundation for understanding how more
+advanced techniques work. This is the reason I will be covering GLM in depth.
+(For those who will go into graduate students or into research in general, you
```
mbonsma: students -> studies

lec11-modelling.Rmd
mbonsma: And also maybe mention that the formula is the equation of a straight line, which is what you're trying to fit to the data.

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+
+$$ y = a + x_1b_1 + x_2b_2 +...+ x_nb_n + e $$
+
+Where each $x_n$ is a variable (either continuous or categorical) in a data frame
+and $b_n$ is the variable coefficient/slope.
+
+We construct these regression models for several reasons. Sometimes we want to
+infer how some variables ($x$) cause or influence another variable ($y$). Or
+maybe we know that $y$ has a lot of error in the measurement or is difficult to
+measure, so we want to derive a formula in order to predict the variable based
```
mbonsma: to predict the variable -> to predict $y$

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+
+Same as above! And for males (equal to 1):
+
+```{r}
+42.17 + (1 * 0.825)
+```
+
+Good, it matches what we calculated. Why is this useful? Well, we can add more
+terms to the GLM equation. This allows us to determine how a variable influences
+the $y$ when other variables are held constant and is known as *multiple linear
```
mbonsma: the $y$ -> $y$

lec11-modelling.Rmd
mbonsma: lol `math doesn't care what is physically possible or not`

and that's how we got black holes. Friggin math.

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+and hindfoot length to zero! But is it physically even possible that a hindfoot be
+non-existant?? No. But math doesn't care what is physically possible or not.
+That's where your ecology/biology knowledge comes in. So let's change that up by
+changing the actual value of hindfoot length in the dataset. There are several
+ways to transform the data into something that is meaningful to us. But first:
+
+> What does everyone think is a meaningful way to transform hindfoot length so
+we can interpret the model more appropriately?
+
+There are many ways to do it, but one way that is often used is to mean center
```
mbonsma, line 190: center -> centre 

Canadian spelling woot woot!

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+complicated than that. So depending on your final project, if you need/want to
+use this, we can help walk you through this more.
+
+## Confidence intervals, and p-values
+
+So what's the point of using `glm`? Not only does it create a model with
+constants and coefficients, but it is also used to calculate how *certain* we
+are about the results. The main power of GLM comes from the ability to estimate
+model parameters and derive meaning from how certain $x$ variables (independent
+variables) influence the $y$ (dependent variable). With his model we can use to
```
mbonsma: We can use this model to predict

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+interval.
+
+```{r}
+fit_hf <- glm(weight ~ hindfoot_length, data = portal)
+tidy(fit_hf, conf.int = TRUE) %>% 
+    select(term, estimate, conf.low, conf.high, p.value)
+```
+
+In this case, the uncertainty of the estimate for hindfoot length is between
+2.524 to 2.58, which is pretty narrow (that's a good thing). This tells us that
```
mbonsma, line 328: Maybe mention which confidence interval you're calculating (i.e. 90%, 95%) and what that means, like 95% means 95% of the probability for that parameter is within that interval.

lec11-modelling.Rmd
```diff
@@ -0,0 +1,527 @@
...
+
+P-values are based on and used in hypothesis testing. Hypothesis testing is when
+you ask a question such as "Are males heavier than females in rodents?". There
+is then the null hypothesis ("females and males weigh the same") and the
+alternative hypothesis ("males are heavier than females"). Using statistical
+tests, we can calculate the p-value to give us an indication of which hypothesis
+is true. Traditionally, if the p-value is <0.05 you 'reject' the null
+hypothesis.
+
+However, the true meaning of the p-value is a bit counter intuitive. The
```
mbonsma: counter-intuitive

mbonsma (approved):
This is very cool! I like the emphasis on real-world consequences for not being a responsible data analyst.

lec11-modelling.Rmd
lwjohnst86: in this case, mean centering is a common statistical term. I tend to think of centre as a place or building, and in stats, center is to the data.

QuLogic: @mbonsma Where were you for the colour/color debate?

mbonsma: Yeah, yeah, it's one thing to use `color` in the code, quite another to spell it the wrong way in the text. lol

QuLogic: Luckily, R supports both, generally...

PR #144, opened 2017-10-11
Included R way over loops (lecture 10)

lwjohnst86:
@mbonsma this would be the R way of doing the nested for loops.


mbonsma:
Nice, this is definitely better.

PR #154, opened 2017-10-19
forgot assignment 7 menu item

lwjohnst86:


lwjohnst86:
plus label position of graphviz image

PR #157, opened 2017-10-21
Final assignment

lwjohnst86:
Thoughts?


assignment-final.Rmd
```diff
@@ -1,5 +1,39 @@
 ---
-title: "Final assignment"
+title: "Final assignment: Scientific report"
+output:
+    html_document:
+        toc: false
 ---
 
-> Uploaded when it is handed out.
+The final assignment is fairly straightforward, but also challenging. You, as a
```
joelostblom: The final assignment is straightforward, but challenging.

assignment-final.Rmd
```diff
@@ -1,5 +1,39 @@
...
+team, must write up a scientific report/manuscript. The best way to think about
```
joelostblom: As a team, you will work together to compile your findings into a scientific report.

assignment-final.Rmd
```diff
@@ -1,5 +1,39 @@
...
+title: "Final assignment: Scientific report"
+output:
+    html_document:
+        toc: false
 ---
 
-> Uploaded when it is handed out.
+The final assignment is fairly straightforward, but also challenging. You, as a
+team, must write up a scientific report/manuscript. The best way to think about 
+this assignment is to imagine you are a scientist and you are writing a paper
```
joelostblom: ~The best way to think about this assignment is to i~Imagine you are scientists ~and you are~ writing a paper to publish your results in a peer reviewed journal.

assignment-final.Rmd
```diff
@@ -1,5 +1,39 @@
...
+to publish your research. This is essentially what we expect of you from the
+final report.
```
joelostblom: ~This is essentially what we expect of you from the final report.~

assignment-final.Rmd
```diff
@@ -1,5 +1,39 @@
...
+
+So, your report should have at least these sections: "Background and Rationale",
```
joelostblom: ~So, y~Your...

assignment-final.Rmd
```diff
@@ -1,5 +1,39 @@
...
+"Research Objectives and Hypotheses", "Methods" (with "Data Description" and
+"Data Analysis" subsections), "Results", "Discussion", and "Final Conclusion".
```
joelostblom: Maybe make a list of the section instead?

assignment-final.Rmd
```diff
@@ -1,5 +1,39 @@
...
+
+For the report, you are expected to:
+
+- Search the previous research and literature on your research questions (this
+doesn't have to be extensive).
```
joelostblom: Yes, I really think this should not be emphasized. Sure they have to write some kind of introduction, but they might spend way to long time on this if we are not clear. And I don't think it matters much if their results are not completely novel, as long as they learn the process of extracting and presenting them.

assignment-final.Rmd
```diff
@@ -1,5 +1,39 @@
...
+
+The report and associated code is expected to:
+
+- Be entirely reproducible.
+- Have well documented code.
+
+You are also expected to:
+
+- Work well as a team.
+- Use Git and GitHub to show your work and workflow with your team.
```
joelostblom, line 45: Paste the grading rubric here also.

I think we should mention something in terms of page length and number of figures. Just to give them a ballpark, what do you think is reasonable?

joelostblom (commented):
Nice, thanks for writing this one!

assignment-final.Rmd
lwjohnst86: hm, didn't think of that. only problem is they might aim for that rather than just write something. But what about... ~2500 words and ~6-8 figures/tables?

joelostblom: Yes, we should stress that we want them to focus on the writing and that we are flexible in these requirements. But a good ballpark is what you suggested I think, ~2500 - 3000 words, and ~6-8 figures (but they are free to group figures together in panels etc), this is more to give them an idea than to use as a hard criteria for submission.

lwjohnst86: @joelostblom updated. Let me know if that is fine.

linamnt:
Should we mention formatting? Maybe we can ask them to pick a journal format and follow those guideleines.
Also, are they to submit as a pdf? Or as RMarkdown again? If they are expected to output as pdf, they'll need to have LaTeX installed, so that might be something we try to get working for the reproducibility lecture, or one of the group work days.

assignment-final.Rmd
```diff
@@ -1,5 +1,61 @@
...
+peer-reviewed journal.
+
+Your report should have at least these sections: 
+
+- "Background and Rationale"
+- "Research Objectives and Hypotheses" 
+- "Methods" (with "Data Description" and "Data Analysis" subsections)
+- "Results"
+- "Discussion"
+- "Final Conclusion"
```
joelostblom: I think this is redundant. I would just call it "Conclusion" or "Conclusions".

assignment-final.Rmd
joelostblom: 👍 Thanks!

joelostblom (approved):

lwjohnst86:
Good point Lina. I think for now we'll keep it simple and just have it a html output. But we can give them the option to make it prettier/as a pdf.

joelostblom:
Thanks @linamnt . I think the default formatting in the R markdown reports looks pretty good. We can experiment with different article formats and adding toc and such if we go the Latex/PDF route. Also colors on heading and links I guess, but let's not get too fancy. I think its great if they can learn how to knit to PDF and we can help them with troubles they run into in class.

PR #172, opened 2017-10-30
Final assignment fix

lwjohnst86:
Mention about coding style + added abstract


joelostblom:
Good idea with being explicit about marking final assignment code style

PR #181, opened 2017-11-09
add lec17 exercise example

linamnt:


joelostblom:
Looks good, will you link this in the lecture file?

PR #182, opened 2017-11-09
link to exercise

linamnt:


lec17-rmarkdown.Rmd
```diff
@@ -384,6 +384,8 @@ kable(head(iris))
 5. Now Knit the file to html and compare with your neighbour to see if you got
 the same output and work together to fix any issues.
 
+Click [here](lec17-exercise.Rmd) for an example of how this file could look.
```
lwjohnst86: You'll have to link to the github blob file, since the rendered lec17 file will be html, not Rmd

PR #193, opened 2018-05-08
Intro edits

aays:
I'm attempting to pare down the first five lessons into just four, so that we can have that extra lecture as discussed last meeting. The extent to which I can do this mostly revolves around what we're electing to keep/remove in the second lecture. I'm not sure how the later lectures are going to pan out from the R end, but if others will be using for loops/vector operations when teaching their material I imagine those same concepts could be introduced then instead of now - thus freeing up even more room in lecture 2 for us to delve into the tidyverse stuff sooner. 

Here's a rough log I kept of the changes I've made so far:

lecture 01
- minor edits and rephrasing in 'working with computers' section
- skipping over tab completion - this will be left for lec 2 in its entirety
- incorporation of assignment + objects here in lec 1 - frees up room in lec 2
- moved section on code chunks from lec 2 to lec 1 when introducing R Notebook
- minor edits/rephrasing in 'saving data and generating reports' section
- tempted to skip over most of the section about yaml headers, save for the PDF setting - I figure this can be talked about later, but leaving it in for now - thoughts?
- are we still doing the background experience survey?

lecture 02
- main goal is to pare down content here and introduce more of lec 3 stuff
- some intro assignment content removed + put into lec 1
- moved writing functions to earlier in the lesson, just after other section on functions
- removed subsetting vectors section - syntax can be introduced as needed
- removed conditional subsetting section - conditions can be introduced alongside`dplyr::filter()`
- removed missing data - can also be introduced with dplyr - `filter(!is.na)`, etc
- added material from lec 3 - introducing dataset, 'what are data frames'

### DIFF
--- a/lec01-introduction.Rmd
+++ b/lec01-introduction.Rmd
@@ -1,6 +1,7 @@
 ---
 title: "Introduction to course"
-author: Joel Östblom
+author: "Joel Östblom & Ahmed Hasan"
+output: pdf_document
 ---
 
 ```{r echo=FALSE, results='hide', message=FALSE, warning=FALSE}
@@ -25,10 +26,11 @@ library(GGally)
 > 
 > Total lesson time: 1 hour
 > 
-> - Working with computers (15 min)
-> - Programming in R (40 min)
+> - Working with computers (10 min)
+> - Programming in R (45 min)
 >     - The Basics (10 min)
 >     - RStudio and the R Notebook (5 min)
+>     - Creating and assigning objects (5 min)
 >     - Plotting (10 min)
 >     - Saving data and generating reports (5 min)
 >     - R Markdown (10 min)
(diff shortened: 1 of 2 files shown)


lec01-introduction.Rmd
```diff
@@ -103,22 +107,18 @@ value of this sequence.
...
 ```
 
-This is not as rich as the English language, which actually is something I am
-grateful for as an international student. There are few exceptions to the
-rules, and you rarely have to remember illogical grammar just because "that is
-how it is", I love it!
-
-And if you think it sounds complicated to remember all of these functions by
-heart, don't worry there are several places to get help. The one I use the most
+<!-- skipping this in lecture since it's covered in lec 2 anyways...
```
joelostblom, line 110: But can you code without tabcompletion? I hit it automatically, so I just explain it here to not drive myself crazy trying to avoid it...

lec02-basic-r.Rmd
```diff
@@ -511,245 +535,347 @@ the hierarchy of how these data types are coerced?
...
+(e.g., characters, integers, factors). We can see this when inspecting the
+structure of a data frame with the function `str()`:
 
 ```{r}
-heights <- c(2, 4, 4, NA, 6)
-mean(heights)
-max(heights)
-mean(heights, na.rm = TRUE)
-max(heights, na.rm = TRUE)
+str(surveys)
```
joelostblom, line 724: Worth mentioning that they see some of this info without `str` for tibbles? Or unnecessary maybe.

joelostblom (approved):
Looks good to me!

lec02-basic-r.Rmd
aays: Perhaps a quick verbal mention - I don't know if it needs to be a separate point here, however. I feel that emphasizing it might confuse students into thinking that printing out the entirety of data frames in general (ie not just tibbles) will display that information.

PR #200, opened 2018-05-25
Remove commented section about tab complete to fix uneven commenting

joelostblom:
Maybe these comments are the reasons why the images are not currently showing up at https://uoftcoders.github.io/rcourse/lec01-introduction.html? Although probably not... any idea why the plot images are not showing up on the site anymore @lwjohnst86 @aays ?

### DIFF
--- a/lec01-introduction.Rmd
+++ b/lec01-introduction.Rmd
@@ -107,19 +107,6 @@ value of this sequence.
 mean(seq(0, 30))
 ```
 
-<!-- skipping this in lecture since it's covered in lec 2 anyways...
-If you think it sounds complicated to remember all of these functions by
-heart, don't worry: there are several places to get help. The one I use the most
-is called tab-complete: often I remember the beginning of the function I want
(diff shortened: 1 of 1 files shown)


joelostblom:
Maybe just something wrong in the last build and we only need totrigger a rebuild and see if it works?

lwjohnst86:
You want my review on the commit or on why the image isn't showing?

lwjohnst86:
I looked into it a bit and I'm not entirely sure why. It could be two things: The code chunk might need to be named (not likely) and maybe when the `output: pdf_document` is added to the top it changes things...?

which by the way, you don't need that output option in the YAML header @aays and @joelostblom

lwjohnst86:
@joelostblom the images work now!

aays:
Good catch @joelostblom - though seems I'm a bit late to the party (thanks for the fix @lwjohnst86 !) 

Might as well get rid of that commented section all the same however, just to keep things clean.

joelostblom:
Great thanks guys!

PR #201, opened 2018-05-25
remove output yaml option

lwjohnst86:
As per @joelostblom question in #200, the reason the image isn't showing is because the YAML header option `output` was set to `pdf_document`... which generates a tex file that R tries to generate into a pdf... but since the travis environment doesn't have TeX installed, it throws an error and doesn't finish creating everything. Not sure why travis didn't catch that...

Anyway, you guys don't need to set the output option in the YAML header.


joelostblom:
Great thank you @lwjohnst86 !!

PR #206, opened 2018-08-28
remove failing ggpairs command

aays:
`ggpairs` is breaking for some inexplicable reason, and I can't seem to debug the error for the life of me - this is interfering with knitting, so I figure we might as well axe it for now


aays:
wait - figured it out! ggpairs lives to see another day

PR #208, opened 2018-08-28
Small typo fix

James-S-Santangelo:


index.Rmd
```diff
@@ -1,5 +1,8 @@
 ---
-title: "Syllabus: EEB313 Quantitative Methods in R for Biology [24L, 12P]"
+title: 'Syllabus: EEB313 Quantitative Methods in R for Biology [24L, 12P]'
+output:
+  pdf_document: default
+  word_document: default
```
lwjohnst86, line 5: Not sure why this has been added. The output doesn't matter since the Rmd is generated into html and the subsequent website.

aays: This was added so that we could knit the syllabus to a separate PDF to be uploaded to Quercus, but now that you mention it I suppose it's not needed on the repo itself.

PR #214, opened 2018-09-17
Minor edits to lec04

aays:
Two small changes I felt needed making as I prepped lecture 4 -

1. Removal of text that mentioned a data visualization lecture later on - no longer the case!
2. 'Tidier' method of creating a subsetted data frame containing most abundant species, which removes the need for an intermediate data frame to be made w/ `tally()` and compared against. Also can be used to demonstrate grouped `mutate` operations

### DIFF
--- a/lec04-dplyr.Rmd
+++ b/lec04-dplyr.Rmd
@@ -636,7 +636,6 @@ surveys_abun_species %>%
 It is possible to specify exactly which colors to use, to avoid those that are
 hard to distinguish. We can also change the thickness of the lines, and adjust
 the x labels so that they don't overlap. 
-data visualization lecture.
 
 ```{r}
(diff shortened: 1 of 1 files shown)


lec04-dplyr.Rmd
```diff
@@ -645,7 +635,7 @@ surveys_abun_species %>%
 
 It is possible to specify exactly which colors to use, to avoid those that are
 hard to distinguish. We can also change the thickness of the lines, and adjust
-the x labels so that they don't overlap. This will be discussed in detail in the
+the x labels so that they don't overlap. 
 data visualization lecture.
```
QuLogic: Did you not also want to remove the end of the sentence?

aays: ...whoops

Good catch!

PR #215, opened 2018-09-19
Adding new lecture 9 — Randomization tests

James-S-Santangelo:
I've adding lecture 9, included the required packages in DESCRIPTION and updated the _site.yml to include a reference to the lecture.


lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,366 @@
...
+    geom_histogram(data = reshuffled[reshuffled$sex == "Male",], aes(x = width), bins = 30, fill = "red", alpha = 0.4, colour = "black") +
+    geom_histogram(data = reshuffled[reshuffled$sex == "Female",], aes(x = width), bins = 30, fill = "blue", alpha = 0.4, colour = "black") +
+    geom_vline(data = reshuffled[reshuffled$sex == "Male",], aes(xintercept = mean(width)), size = 1, linetype = "dashed", colour = "red") +
+    geom_vline(data = reshuffled[reshuffled$sex == "Female",], aes(xintercept = mean(width)), size = 1, linetype = "dashed", colour = "blue") +
+    theme_classic()
+
+# Let get the mean difference of this single simulation
+means_sim1 <- group_by(reshuffled, sex) %>% summarize(mean = mean(width))
+mean_diff_sim1 <- means_sim1[means_sim1$sex == "Male", "mean"] - means_sim1[means_sim1$sex == "Female", "mean"]
+cat("The mean difference from a single reshuffling of the data is", mean_diff_sim1$mean)
```
lwjohnst86: This caused the travis fail.

James-S-Santangelo:
Hmm still failing in the next code chunk despite no issues on my end.

James-S-Santangelo:
So it knits without issue on my end and the previous knits to HTML through Travis were fine. For some reason it's failing now due to a call to `select()` that's not even present in the code chunk.

lwjohnst86:
I can't find where it's going wrong. But my guess is one of the underlying
packages isn't using select properly. Try loading dplyr at the very top of
the file.

James-S-Santangelo:
Still nothing. I think the issue is a conflict between `dplyr::select()` and `MASS::select()`. The `EcoSimR` package that I'm using requires `MASS` and that is where the code is breaking. However, loading `dplyr` after `EcoSimR` or vice-versa both lead to the same error.

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+> ### Setup
+>
+> - `install.packages('dplyr')` (or `tidyverse`)
+> - `install.packages('ggplot2')` (or `tidyverse`)
+> - `install.packages("reshape2")`
+> - `install.packages("EcoSimR")`
+
+## Introduction to randomization tests
+
+As we saw in previous weeks, standard parametric statistical tests (e.g. ANOVA, standard linear regression) have numerous assumptions that should be respected prior to drawing conclusions from their results. Violation of these assumtions can lead to erroneous conclusions about the population(s) you are trying to understand. Mixed-effects models provided us with a few ways of overcoming the limitations of standard models, while still using all of our data and remaining in a statistical framework that we generally understand and are comfortable with. Randomization tests provide yet another alternative solution to paramateric tests.
```
QuLogic: s/assumtions/assumptions/; s/paramateric/parametric/ (I think)

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+
+## An aside on simulating data in R
+
+R allows us to genetate fake data (i.e. simulate data) rather easily. This is useful if we just want some toy data to play around with, want to simulate data simular to the type we will collect in an experiment to test our analyses, or want to simulate data collection many times to estimate population parameters (e.g. variance or standard deviation). We can sample existing (i.e. real) data using the `sample()` function, or take random samples from known statistical distributions (e.g. Normal distribtuion). We will return to `sample()` later in this lesson. For now, let's focus on sampling from known distributions.
```
QuLogic: s/genetate/generate/; s/simular/similar/; s/distribtuion/distribution/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+```{r}
+ggplot() +
+    ylab("Count") + xlab("Simulated mean difference") +
+    geom_histogram(aes(x = simulated_means), bins = 30, fill = "grey", alpha = 0.4, colour = "black") +
+    geom_vline(xintercept = diff_means_obs, size = 1, linetype = "dashed", colour = "black") + 
+    theme_classic()
+
+```
+
+Finally, to get out P-value, we calculate the number of times  the simulated mean difference exeeded the observed mean difference from our data. Because we are performing a two-tailed test, this amounts to determining the number of times the simulated mean difference is **either greater or lesser than the observed difference**. We can do this by asking how many times the absolute value of the simulated mean difference is greater or equal to the absolute value of the observed mean difference.
```
QuLogic: s/exeeded/exceeded/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+
+```{r}
+abs_simulated_means <- abs(simulated_means)
+abs_diff_means_obs <- abs(diff_means_obs)
+exceed_count <- length(abs_simulated_means[abs_simulated_means >= abs_diff_means_obs])
+p_val <- exceed_count / nreps
+cat("The P-value from the randomization test is", p_val)
+```
+
+As we can see, the randomization test provides results that are largely consistent with the t-test. This is not surprizing since we specifically sampled independent data from normal distributions with similar variances. In other words, our sampled data do not violate any of the assumptions of the t-test and in such a case the difference in means is by definition analogous to the _t_ statistic used in the t-test. There are some cases where randomization tests provide more accurate (or _exact_) P-values than parametric tests (e.g. small samples from skewed distributions) but even more diverse applications of randomization tests are found in ecology and evolutionary biology.
```
QuLogic: s/surprizing/surprising/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+
+```{r}
+# Load in example dataset
+comMatrix_example <- read.csv("data/lec09_CommunityMatrix_Example.csv")
+head(comMatrix_example)
+colSums(comMatrix_example[,2:5]) # Sliced to return a numeric Matrix, which is required for colSums() and rowSums()
+rowSums(comMatrix_example[,2:5])
+```
+
+As you can see, the data is a matrix with sites (e.g. islands, quadrats, etc.) as columns and species as rows. Each cell represents whether a species is present (1) or absent (0) at that site. This type of data is often referred to as a _presence-absence matrix_. The column totals (using `colSums()`) gives us a sense of how species richness varies across the sites while the row totals (using `rowSums()`) tells us how rare or common a species is across all sites. The question we are interested in addressing is: **Are the observed species co-occurence patterns the result of species interactions or random chance?**
```
QuLogic: s/co-occurence/co-occurrence/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+neon_data_filtered <- unique(neon_data_filtered) # Keep only unique rows so that cells represent presence-absence and not abundances
+PA_matrix <- dcast(neon_data_filtered, formula = taxonID ~ plotID, fun.aggregate = length)
+head(PA_matrix)
+```
+
+Well that was painless! We now have our data in a form that is usable in our randomization test but we need to decide two things first: (1) _What is our test statistic?_ This is what we'll use to determine if co-occurence patterns are random or not, similar to how we used the difference in means to tell if male and female beetles differed in length. (2) _How do we randomize such a large matrix?_ Do we randomly permute rows and columns so that each cell is equally likely to contain a 1 or 0? Do we constrain row permutations so that row totals (i.e. species rarity) must be preserved but site richness is free to vary? Do we constrain column permutations so that column totals (i.e. site richness) must be presereved but species rarirty can vary? It turns out there are exactly 9 different ways a matrix like this can be randomized, each involving a different set of constraints and biological assumptions (Gotelli 2000).
```
QuLogic: s/presereved/preserved/; s/rarirty/rarity/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+
+For our purpose, we will preserve species rarity but will assume that each site is equally likly to be colonized. In other words, we are assuming that sites do not vary in habitat quality or anything else that may change the number of species that can occupy it. However, we want rare species to remain rare throughout the permutations. This model is analogous to the _SIM2_ algorithm in Gotelli (2000) and is reasonable in our case since the sites are relatively close in proximity (i.e. likely have similar habitat quality).
```
QuLogic: s/likly/likely/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+
+For our purpose, we will preserve species rarity but will assume that each site is equally likly to be colonized. In other words, we are assuming that sites do not vary in habitat quality or anything else that may change the number of species that can occupy it. However, we want rare species to remain rare throughout the permutations. This model is analogous to the _SIM2_ algorithm in Gotelli (2000) and is reasonable in our case since the sites are relatively close in proximity (i.e. likely have similar habitat quality). 
+
+For our index, we will use the **C-score** (Stone and Roberts 1990). I will not go into the calculation of this score here but suffice it to say that the higher the C-score, the less co-occurence (on average) between all pairs of species in the matrix (i.e. a more seggregated matrix). Smaller C-scores indicate an aggregated matrix (i.e. more co-occurence among species pairs). Thankfully for us, there is an R package that implements the randomization algorithm for us. Let's go ahead and conduct the analysis.
```
QuLogic: s/co-occurence/co-occurrence/g;  s/seggregated/segregated/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+
+```{r message=FALSE, warning=FALSE}
+set.seed(50)
+
+# Perform co-occurence analysis
```
QuLogic: s/co-occurence/co-occurrence/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+co_oc_analysis <- cooc_null_model(PA_matrix, algo = "sim2", 
+                                  nReps=1000, 
+                                  metric = "c_score", 
+                                  suppressProg = TRUE, 
+                                  saveSeed = TRUE)
+
+
+# Summarize output from co-occurence analysis
```
QuLogic: s/co-occurence/co-occurrence/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+summary(co_oc_analysis)
+
+# Let plot the randomization test
+plot(co_oc_analysis,type="hist")
+```
+
+As we can see, our observed C-scrore (red line) is much lower (significantly so) than we would expect by random chance given the species in our community. This suggests the community is _aggregated_ such that some species are more likely to occur with other species in the community than we would expect if the communities were assembling themselves randomly. This can happen if there is facilitation and some species create ecological niches or favourable habitat for other species such that they occur more often. Importantly, the C-score is an average across all pairwise comparisons in the matrix and thus doesn't tell us anything about which specific species pairs may be more or less likely to co-occur. Identifying which species pairs are driving this pattern would require quite a bit more time and identifying the ecological mechanisms behind co-occurence patterns of specific species pairs requires careful experimentation.
```
QuLogic: s/C-scrore/C-score/ (I think); s/co-occurence/co-occurrence/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+
+```{r}
+plot(co_oc_analysis,type="cooc")
+```
+
+Notice how all of the rows in the two figures above always have the same number of filled squares. This is confirmation that the rarity of species (i.e. row totals) has been preserved during the permutation. 
+
+## Other uses of randomization tests
+
+Testing for mean differences among groups and for putative rules governing community assembly are only two of the many uses for randomization tests. In this section, I brieflly highlight other uses of data randomization in statistics, ecology, and evolutionary biology.
```
QuLogic: s/brieflly/briefly/; s/randmonization/randomization/

lec09-Randomization-tests.Rmd
```diff
@@ -0,0 +1,364 @@
...
+
+**Monte-carlo methods** are a huge class of methods that generally involve simulating random numbers to estimate probability distributions and for obtaining numerical solution to really complicated problems. For our purposes, these are especially common in computational biology, genomics and in Bayesian inference (e.g. building phylogenies).
+
+## Additional reading
+
+1. Diamond, J.M. 1975. Assembly of species communities. p. 342-444 in: Ecology and Evolution of Communities. M.L. Cody and J.M. Diamond (eds.). Harvard University Press, Cambridge.
+2. Connor, E.F. and D. Simberloff. 1979. The assembly of species communities: chance or competition? Ecology 60: 1132-1140.
+3. Connor, E.F., M.D. Collins, and D. Simberloff. 2013. The checkered history of checkerboard distributions. Ecology 94: 2403-2414
+4. Stone. L. and A. Roberts. 1990. The checkerboard score and species distributions. Oecologia 85: 74-79.
+5. Manly, B. 2007. Randomization, bootstrap and Monte Carlo methods in biology. B. Carlin, C. Chatfield M. Tanner andJ. Zidek, eds. 3rd edition. Chapman and Hall/CRC.
```
QuLogic: s/andJ/and J/

QuLogic (commented):
Some typos:

QuLogic:
Are you building the single file locally (in R studio or `rmarkdown::render_site('lec09-Randomization-tests.Rmd')`) or the whole site like Travis does (with `rmarkdown::render_site('.')`)? Because I can reproduce the problem if I build the whole site, but not if I just build the new lecture individually.

QuLogic:
Well, I managed to hack a fix by adding this to the end of the lec09 file:
```r
unloadNamespace("lmerTest")
unloadNamespace("lme4")
unloadNamespace("EcoSimR")
unloadNamespace("MASS")
```
but that's somewhat annoying.

(also, I realized that my previous question is moot since it's lec 10 that's failing.)

QuLogic:
There's an rmarkdown bug for this rstudio/rmarkdown#1326 and it will be fixed by setting `new_session: true` with 1.11, whenever that's released.

lwjohnst86:
Yes it is a bit hacky but the unloadNamespace works well enough. Maybe do
it before using ecosimr and then load dplyr again after? You can use hidden
code chunks with these functions so it removes it from the website output.

QuLogic:
No, it needs to go at the end of lecture 9; the failure is actually in lecture 10.

James-S-Santangelo:
Thank for all the help everyone! I'll push the changes shortly. I assumed the error was in lecture 9 since it never occurred in any previous PRs but you're right that it's in 10. Weird that my lecture 9 would cause lecture 10 to break. I didn't realize the namespace is maintained across lectures in that way.

James-S-Santangelo:
Victory! Thanks all!

PR #226, opened 2018-09-25
Fix mice code and formatting in lecture 6

QuLogic:
Headings need a space after `#`. The broken mice code was a small typo: `ppm` -> `pmm`.

I don't think the extra backticks did any harm; this is just for consistency.

### DIFF
--- a/lec06-exploratory-data-analysis.Rmd
+++ b/lec06-exploratory-data-analysis.Rmd
@@ -250,7 +250,7 @@ To impute the missing values, MICE package uses an algorithm in a such a way tha
 
 The code below will remove a variable as a predictor in imputing missing values, but will not stop MICE from imputing values for that variable itself.
 ```{r}
-predM[, c("record_id")]=0
+predM[, c("record_id")] <- 0
 ```
 
 By default mice() includes all the variables in the data; it is suggested that we use as many variables as possible. Do not exclude a variable as predictor from the imputation model if that variable is going be included in your final analysis model.
@@ -258,7 +258,7 @@ By default mice() includes all the variables in the data; it is suggested that w
 
 If you want to skip a variable from imputation itself use the code below. This variable will still be used for prediction. Let's exclude all the variables where we aren't missing data or where imputation isn't of interest:
 ```{r}
-meth[c("year", "day", "month", "plot_id", "species_id", "plot_type", "genus", "species", "taxa")]=""
+meth[c("year", "day", "month", "plot_id", "species_id", "plot_type", "genus", "species", "taxa")] <- ""
 ```
 
 Now let specify the methods for imputing the missing values. There are specific methods for continuous, binary and ordinal variables. We will now set different methods for each variable. You can add more than one variable in each method. The default for continuous data is pmm. pmm stands for predictive mean matching, the default method of `mice()` for imputation of continous incomplete variables; for each missing value, pmm *finds a set of observed values with the closest predicted mean as the missing one and imputes the missing values by a random draw from that set*. Therefore, pmm is restricted to the observed values, and might do fine even for categorical data (though not recommended).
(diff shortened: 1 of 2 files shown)


QuLogic:
Ugh, stupid global namespace.

lwjohnst86:
"stupid global namespace" I think it's less about namespace and more about improper package creation by authors of mice etc. This is an area I think Python is really good at... since you need to specify the package that the function is coming from (`npm.function_name`) so you would never encounter this problem... but R doesn't have this. Unfortunate :(

lec06-exploratory-data-analysis.Rmd
```diff
@@ -210,73 +210,73 @@ ggplot(data = diamonds2, aes(x = x, y = y)) +
...
+meth <- init$method
+predM <- init$predictorMatrix
 ```
 
 To impute the missing values, MICE package uses an algorithm in a such a way that information from all the other variables in the dataset is used to predict and impute the missing values. Therefore, you may not want to use a certain variable as a predictor. For example, the ID variable does not have any predictive value, so we want to remove it. 
 
 The code below will remove a variable as a predictor in imputing missing values, but will not stop MICE from imputing values for that variable itself.
 ```{r}
-# predM[, c("record_id")]=0
+predM[, c("record_id")]=0
```
lwjohnst86: = to <-

lec06-exploratory-data-analysis.Rmd
```diff
@@ -210,73 +210,73 @@ ggplot(data = diamonds2, aes(x = x, y = y)) +
...
 ```
 
 By default mice() includes all the variables in the data; it is suggested that we use as many variables as possible. Do not exclude a variable as predictor from the imputation model if that variable is going be included in your final analysis model.
 
 
 If you want to skip a variable from imputation itself use the code below. This variable will still be used for prediction. Let's exclude all the variables where we aren't missing data or where imputation isn't of interest:
 ```{r}
-# meth[c("year", "day", "month", "plot_id", "species_id", "plot_type", "genus", "species", "taxa")]=""
+meth[c("year", "day", "month", "plot_id", "species_id", "plot_type", "genus", "species", "taxa")]=""
```
lwjohnst86: <-

lec06-exploratory-data-analysis.Rmd
```diff
@@ -210,73 +210,73 @@ ggplot(data = diamonds2, aes(x = x, y = y)) +
...
 ```{r}
-# init$meth
+init$meth
 ```
 
-The variables ``hindfoot_length`` and ``weight`` are already set to ppm, so we will leave it like that. However, if we wanted to directly specify the method, we could use the following code:
+The variables `hindfoot_length` and `weight` are already set to pmm, so we will leave it like that. However, if we wanted to directly specify the method, we could use the following code:
 ```{r}
-# meth[c("hindfoot_length")]="ppm"
+meth[c("hindfoot_length")]="pmm"
```
lwjohnst86: <-

lec06-exploratory-data-analysis.Rmd
```diff
@@ -210,73 +210,73 @@ ggplot(data = diamonds2, aes(x = x, y = y)) +
...
 ```
 
 Now it is time to run the multiple (m=5; this is generally standard) imputation.
 ```{r}
-# set.seed(103)
-# imputed = mice(survey, method=meth, predictorMatrix=predM, m=5)
+set.seed(103)
+imputed = mice(survey, method=meth, predictorMatrix=predM, m=5)
```
lwjohnst86: <-

QuLogic:
I mean the global namespace when rmarkdown generates a site, and keeps the namespace around for subsequent files.

lec06-exploratory-data-analysis.Rmd
```diff
@@ -286,36 +286,40 @@ method – Refers to method used in imputation. We used predictive mean matching
...
+set.seed(10)
+iris2[sample(1:nrow(iris2), 20), "Sepal.Length"] <- NA
+iris2[sample(1:nrow(iris2), 20), "Petal.Length"] <- NA
 ```
 
-
+```{r, show=FALSE}
+unloadNamespace("mice")
+unloadNamespace("mitml")
+unloadNamespace("jomo")
```
lwjohnst86, line 324: just remove this line

lwjohnst86 (changes requested):

lec06-exploratory-data-analysis.Rmd
QuLogic: Can't; [breaks lec9](https://travis-ci.org/UofTCoders/rcourse/builds/433090310#L2870).

PR #230, opened 2018-09-26
Removing messages from Tidyverse import 

James-S-Santangelo:
Also eliminating question text that is included in an indented code chunk.

### DIFF
--- a/assignment-04.Rmd
+++ b/assignment-04.Rmd
@@ -14,7 +14,7 @@ you assignment as a knitted `.pdf` (prefered) or `.html` file.*
 1. Visualization (2 marks)
 Import the tidyverse library. We will be using the same beaver1 dataset that we used in last week's assignment. 
 
-```{r}
+```{r message=FALSE, warning=FALSE}
 library(tidyverse)
 ```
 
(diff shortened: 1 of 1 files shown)


assignment-04.Rmd
```diff
@@ -45,14 +46,17 @@ a. Look through the help documentation (?CO2) to understand what each variable m
...
-    Use logistic regression to see if weight significantly predicts sex. Make a              concluding statement as to indicate whether the model is significant and create a        plot to visualize the linear model. Run the following code to ensure sex is treated      as a factor variable:
+
+Use logistic regression to see if weight significantly predicts sex. Make a concluding statement as to indicate whether the model is significant and create a plot to visualize the linear model. Run the following code to ensure sex is treated as a factor variable:
+
 ```{r}
 survey$sex <- as.factor(survey$sex)
 ```
-      Hint: you need to make sure there are only two levels to this variable: ``"F"`` and       ``"M"``. (0.5 marks)
+
+Hint: you need to make sure there are only two levels to this variable: ``"F"`` and ``"M"``. (0.5 marks)
```
lwjohnst86: Just fyi, you don't need two backticks for inline code highlighting, you only need one.

PR #236, opened 2018-09-28
Minor revisions to lectures 8 and 9

James-S-Santangelo:


James-S-Santangelo:
Only because the URLs are long and will overshoot the margins in the compiled pdf. I opted to `paste` the URL together over multiple lines and then download the data from the URL. Is that what you're talking about or do you mean the comments? Either way I'm open to suggestions.

lec09-Randomization-tests.Rmd
```diff
@@ -144,10 +146,16 @@ df_females <- data.frame(
...
+    geom_histogram(data = df_males, aes(x = width), 
+                   bins = 30, fill = "red", alpha = 0.4, 
+                   colour = "black") +
+    geom_histogram(data = df_females, aes(x = width), 
+                   bins = 30, fill = "blue", alpha = 0.4, 
+                   colour = "black") +
+    geom_vline(data = df_males, aes(xintercept = mean(width)), 
+               size = 1, linetype = "dashed", colour = "red") +
+    geom_vline(data = df_females, aes(xintercept = mean(width)), 
+               size = 1, linetype = "dashed", colour = "blue") +
```
lwjohnst86, line 156: You know, instead of having two datasets to use to make two plots, you can simplify the code substantially by using one dataset with both male and female. Then using `colour` aes or `group` or even `facet_grid`.  That way you have way less code to deal with. For instance:

```
ggplot(df, aes(x = width, fill = Sex, colour = Sex) +
  geom_histogram(bins = 30, colour = "black", alpha = 0.4) +
  geom_vline(aes(xintercept = mean(width)), size = 1, linetype = "dashed")
```

lec09-Randomization-tests.Rmd
```diff
@@ -162,7 +170,8 @@ mean_males <- mean(df_males$width)
 mean_females <- mean(df_females$width)
 
 diff_means_obs <- mean_males - mean_females
-cat("The observed difference in mean body width between male and female Hercules beetles is", diff_means_obs)
+cat("The observed difference in mean body width between 
+    male and female Hercules beetles is", diff_means_obs)
```
lwjohnst86: Any reason you use `cat` versus using inline R code? Like so:

```
The observed difference in mean body width between male 
and female Hercules beetles is `r diff_means_obs`.
```

This way you get the nicely rendered text, unlike `cat` which prints as a text code chunk.

lec09-Randomization-tests.Rmd
```diff
@@ -195,17 +205,26 @@ Note the differences in body width values between the original and reshuffled da
...
+                   alpha = 0.4, colour = "black") +
+    geom_histogram(data = reshuffled[reshuffled$sex == "Female",], 
+                   aes(x = width), bins = 30, fill = "blue", 
+                   alpha = 0.4, colour = "black") +
+    geom_vline(data = reshuffled[reshuffled$sex == "Male",], 
+               aes(xintercept = mean(width)), size = 1, 
+               linetype = "dashed", colour = "red") +
+    geom_vline(data = reshuffled[reshuffled$sex == "Female",], 
+               aes(xintercept = mean(width)), size = 1, 
+               linetype = "dashed", colour = "blue") +
```
lwjohnst86: This is even more set up to use `fill = Sex`, since `reshuffled` is already one dataset! Why not use that instead of this more complicated code?

lec09-Randomization-tests.Rmd
```diff
@@ -276,9 +300,16 @@ First, let's look at the type of data we need for this analysis.
 
 ```{r}
 # Load in example dataset
-comMatrix_example <- read.csv("data/lec09_CommunityMatrix_Example.csv")
+comMatrix_example <-paste0("https://raw.githubusercontent.com/",                                         "James-S-Santangelo/rcourse/master/data/",
```
lwjohnst86: I assume you didn't mean for the many spaces between commas here.

James-S-Santangelo:
Re: use of `cat`. You're right. Inline code would be much nicer. 

Re: `ggplot`. Yeah good call. I copied this code from some that I wrote a long time ago. Will definitely simplify it. 

Re: Spaces after commas. No I didn't mean for this.

lec09-Randomization-tests.Rmd
```diff
@@ -53,7 +53,7 @@ library(dplyr)
 library(EcoSimR)
 ```
 
-```{r}
+```{r message=FALSE, warning=FALSE}
```
lwjohnst86, line 56: Another suggestion. If you want to have `message=FALSE, warning=FALSE` for all code chunks, you can do that in one line. By including the below code in the very first code chunk at the top, you can have the chunk option set globally for the document:

```
knitr::opts_chunk$set(warning = FALSE, message = FALSE)
```

James-S-Santangelo: Oh sweet I didn't know that! Thanks for the heads up.

PR #238, opened 2018-10-01
Changes to assignment 2

aays:
Turns out actually marking assignment 2 made me notice a few things worth changing for next time:

- best to avoid naming a variable`c` - some students were confused about the potential naming clash with `c()`
- few typo fixes
- the question 3 mark breakdown added up to 2.5 marks, not 3 marks

### DIFF
--- a/assignment-02.Rmd
+++ b/assignment-02.Rmd
@@ -7,7 +7,7 @@ output:
 
 *To submit this assignment, upload the full document on blackboard,
 including the original questions, your code, and the output. Submit
-you assignment as a knitted `.pdf` (prefered) or `.html` file.*
+you assignment as a knitted `.pdf` (preferred) or `.html` file.*
 
 1.  Variable assignment (1 mark)
 
--- a/assignment-02.Rmd
+++ b/assignment-02.Rmd
@@ -33,8 +33,8 @@ you assignment as a knitted `.pdf` (preferred) or `.html` file.*
     b.  What is the difference in lengths of the vectors `v` and `w`?
         (0.25 marks)
 
-    c.  Create a new vector, `v_square`, with the square of elements 3,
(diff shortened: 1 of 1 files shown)


assignment-02.Rmd
```diff
@@ -17,12 +17,12 @@ you assignment as a knitted `.pdf` (prefered) or `.html` file.*
...
         (0.25 marks)
 
-    c.  Assign the product of `a` and `b` to `c`. Display `c`. (0.25
+    c.  Assign the product of `a` and `b` to `product`. Display `product`. (0.25
         marks)
 
     d.  Write a function that adds two numbers and returns their sum.
-        Use it to assign the sum of `a` and `b` to `c`. Display `c`. (In
-        practice, there is already a more sophisticated built-in
+        Use it to assign the sum of `a` and `b` to `product`. Display `product`.
```
QuLogic: But the sum of two things is not a product... Maybe name it `result` instead?

aays: *head meets desk*

assignment-02.Rmd
```diff
@@ -17,12 +17,12 @@ you assignment as a knitted `.pdf` (prefered) or `.html` file.*
...
-    c.  Assign the product of `a` and `b` to `c`. Display `c`. (0.25
+    c.  Assign the product of `a` and `b` to `product`. Display `product`. (0.25
         marks)
 
     d.  Write a function that adds two numbers and returns their sum.
-        Use it to assign the sum of `a` and `b` to `c`. Display `c`. (In
-        practice, there is already a more sophisticated built-in
+        Use it to assign the sum of `a` and `b` to `result`. Display `result`.
+        (In practice, there is already a more sophisticated built-in
         function for this: `c <- sum(a, b)`) (0.25 marks)
```
QuLogic: This `c` should probably also be `result`.

PR #241, opened 2018-10-11
lecture 12 - numerical models

mbonsma:


lec12-numerical-models.Rmd
```diff
@@ -0,0 +1,902 @@
...
+
+$\Delta N$, the change in the population size, will depend on the specifics of our model. Let's rearrange this
+expression into the definition of a derivative by moving $N_t$ over and dividing both sides by $\Delta t$,
+then taking the limit of $\Delta t$ going to $0$:
+
+$$\frac{N_{t+\Delta t} - N_t}{\Delta t} = \frac{\Delta N}{\Delta t}$$
+$$lim_{\Delta t \to 0}\frac{N_{t+\Delta t} - N_t}{\Delta t} = lim_{\Delta t \to 0}\frac{\Delta N}{\Delta t} = \frac{dN}{dt}$$ 
+
+If you've taken calculus, you will recognize this as the first derivative of $N$ with respect to $t$. 
+If your calculus is foggy in your memory, when you see this notation think "change in $N$ per change in $t$.
```
James-S-Santangelo: Do you think it's worth relating this back to what we taught them about the interpretation of beta coefficients from linear models? (i.e. the change in `y` for every one unit change in `x`.

lec12-numerical-models.Rmd
```diff
@@ -0,0 +1,902 @@
...
+
+Here's an example from an earlier lecture:
+
+```{r}
+# For each number in v, print the number.
+v <- c(2, 4, 6)
+for (num in v) {
+    print(num)
+}
+```
```
James-S-Santangelo, line 677: Just FYI, I had them do `for loops` in lecture 9 and assignment 5. Nonetheless I'm sure a refresher will be appreciated.

lec12-numerical-models.Rmd
```diff
@@ -0,0 +1,902 @@
...
+
+## Analyzing models
+
+When you analyze a model, the exact questions you answer will depend on the model. 
+But in general, the following questions are ones you will try to answer for any population model.
+
+1. What are the **fixed points** of the model? (For which sizes of the population(s) does the population size remain constant in time?) 
+2. Which of the fixed points are **stable**? (If the population is slightly different from the fixed point, will it go towards it or away from it?)
+3. How do the above depend on parameters of the model?
+4. Which parameters fit my data best, and how good is the fit?
```
James-S-Santangelo, line 198: Doesn't look like you go into detail here, but I wonder how this relates to the model fitting and selection we did in lecture 11. Might be a nice complement to think about (not in this lecture necessarily). Just curious.

lec12-numerical-models.Rmd
mbonsma: Good idea!

lec12-numerical-models.Rmd
mbonsma: :+1:

lec12-numerical-models.Rmd
mbonsma: Yeah, I was trying to think how to compare these kinds of fits with the stuff you covered. I'm not going into any detail on Maximum Likelihood, but in theory you can just use ML to do the fitting I'm talking about and then calculate AIC as usual. I'll try to work up an example just for completeness, but probably won't cover it in class.

PR #259, opened 2018-10-22
Document outlining project deliverables

mbonsma:
We can upload this as a pdf on Quercus. It contains:

* A description of the project and project goals
* Guidelines for choosing their own data if they want to
* Detailed project deliverables
* Grading rubric

We still need to add the link to the datasets, I'm waiting on that because we'll be updating the name of the datasets lecture.

### DIFF
--- a/project_info.Rmd
+++ b/project_info.Rmd
@@ -24,7 +24,7 @@ that you may not submit anything twice: any work you do as part of this course m
 be submitted for credit in another course (such as a fourth-year research project) and vice versa. 
 If choosing a dataset not listed, make sure it is well-documented, legitimate, and 
 complex enough to support your analysis efforts. Your work should be original; your
-project should not be a reproduction of published analysis. 
+project should not be a reproduction of published analyses. 
 
 ### Project deliverables
 
--- a/project_info.Rmd
+++ b/project_info.Rmd
@@ -1,0 +1,73 @@
+---
+title: "Final project information"
+output:
+  html_document: default
+  html_notebook: default
+---
+
+## Project description
+
+The course project is a self-directed group data analysis project using real 
+ecological data and rigorous scientific methods. Groups are expected to hypothesize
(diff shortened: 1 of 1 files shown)


project_info.Rmd
```diff
@@ -0,0 +1,82 @@
...
+
+### Data
+
+A list of recommended datasets can be found [here](). You are welcome to choose a
+dataset not listed, or data collected as part of a research project, but keep in mind
+that you may not submit anything twice: any work you do as part of this course may not
+be submitted for credit in another course (such as a fourth-year research project) and vice versa. 
+If choosing a dataset not listed, make sure it is well-documented, legitimate, and 
+complex enough to support your analysis efforts. Your work should be original; your
+project should not be a reproduction of published analysis.
```
James-S-Santangelo: Change 'analysis' to 'analyses'

project_info.Rmd
```diff
@@ -0,0 +1,82 @@
...
+encouraged to publish their work on [figshare](https://figshare.com/), an open, 
+citable repository of scientific content. 
+
+All items (except the presentation) are due on December 5th at 11:59 pm.
+
+## Project submission
+
+The project report and code should be submitted on GitHub - each group will have 
+their own GitHub repository to which you can upload your report and code. 
+
```
James-S-Santangelo, line 70: Should we mention something about the format for the final report? For example, should they have Rmd and pdf files on their GitHub or will we accept word documents?

project_info.Rmd
mbonsma:
```suggestion
project should not be a reproduction of published analyses. 
```

project_info.Rmd
mbonsma: That's something I want to discuss at our meeting - we want it to be reproducible, so should we make them write the whole report as an R markdown document? Then they would submit pdf or html plus .Rmd.

PR #260, opened 2018-10-22
Rename lec12 to lec14

mbonsma:


QuLogic:
This doesn't seem to be triggering Travis at all; can you open an entirely new PR?

PR #262, opened 2018-10-23
Combine new project_updates file with assignment_final

mbonsma:


mbonsma:
This one is apparently failing because we've hit a GitHub API rate limit:

```
Error: HTTP error 403.
  API rate limit exceeded for 147.75.198.203. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
  Rate limit remaining: 0/60
  Rate limit reset at: 2018-10-23 16:01:59 UTC
  To increase your GitHub API rate limit
  - Use `usethis::browse_github_pat()` to create a Personal Access Token.
  - Add `GITHUB_PAT` to your travis settings as an encrypted variable.
```

PR #265, opened 2018-10-24
TOKEN -> PAT

lwjohnst86:
seeing if this will increase the API limit


mbonsma:
:pray: thank you @lwjohnst86!

PR #270, opened 2018-10-29
Add assign08

James-S-Santangelo:
Adding updated version of assignment 8.

+ This assignment is currently the same as last year's assignment 7
+ I have changed the assignment to reflect the fact that the project repos are now in the eeb313-2018 organization rather than UofTCoders.

### DIFF
--- a/assignment-08.Rmd
+++ b/assignment-08.Rmd
@@ -1,80 +1,102 @@
 ---
-title: 'Assignment: Data visualization (8 marks)'
+title: 'Assignment 7: Project and team setup (8 marks)'
 output:
     html_document:
         toc: false
 ---
 
-<!-- keep this part -->
 *To submit this assignment, upload the full document on blackboard,
-including the original questions, your code, and the output. Submit both
-your `.Rmd` source file and a knitted `.html` version of the same file.*
+including the original questions, your code, and the output. Submit
+your assignment as a knitted `.pdf` or `.html` file.*
 
-1.  Recreate a customized plot using what you've learned (3.5 marks).
+These assignment tasks seem like a lot of work. However, we will be
+doing these tasks already during class and during project work. The
+purpose of the assignment is for you to get something started on your
+final project. While most of these assignment tasks are group based,
+please fill out each as an individual. In particular, the first question
+should be of *your* forked repository, not one of your team members.
+Each team member should submit their *own assignment*.
 
-    We will use the CO2 dataset studying the cold tolerance on the grass
-    species *Echinochloa crus-galli*.
-
-```{r}
-# Load the CO2 dataset in base R (there is also a co2 dataset, be careful with capitalization)
-CO2 <- datasets::CO2
-```
-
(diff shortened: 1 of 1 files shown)


aays:
@James-S-Santangelo Looks great overall! Two quick things:

1. Most of the students have already forked their repo in class. I'm not sure whether the wording here has to be changed as a consequence (I don't think it makes a huge difference) but just thought I'd point that out all the same.
2. Question 4 presupposes that there is already a `doc` folder - it may be helpful to word the question as 'create this folder and then do x and y things' to avoid confusion. Also, since I didn't get to cover this in class (my bad!) perhaps it'd be a good idea to add a hint to the question containing a link to [this guide](https://github.com/KirstieJane/STEMMRoleModels/wiki/Creating-new-folders-in-GitHub-repository-via-the-browser) on how to create folders in the GitHub editor.

assignment-08.Rmd
```diff
@@ -1,80 +1,102 @@
...
+            Forstmann BU, et al. (2014) Data from: Multi-modal ultra-high resolution
+            structural 7-Tesla MRI data repository. Dryad Digital Repository.
+            (https://dx.doi.org/10.5061/dryad.fb41s)
+
+    -   Fill out the "Team description" section by writing down a brief
+        biography of each member (including what their skills are and
+        what their approximate responsibilities (which can change later)
+        are for the project) and how team roles will be rotated.
+
+    -   URL of team member's pull request for `README.md`:
```
mbonsma, line 74: Should we make each of them do an individual pull request for their bio section? Or not, in case it generates conflicts?

assignment-08.Rmd
```diff
@@ -1,80 +1,102 @@
...
+
+4.  Create a file in the `doc/` folder called `objectives.md`. Create
```
mbonsma: Revise this like you suggested @James-S-Santangelo to 1) make a `doc/` folder and 2) make a file called `objectives.md`.

mbonsma:
I agree with your suggestions. I think it's fine that they already forked that repo, we don't need to change that question.

Is this assignment too group-based? Should we make them do some individual stuff as well?

Are the documents they're making consistent with what's needed for an R project? Was that talked about in class at all?

James-S-Santangelo:
I think these documents are consistent with Rprojects, aside from actually having a .Rproj file. Do we want to have them create the Rproject as part of this assignment?

aays:
R projects weren't covered in class, unfortunately - I had just about enough time to cover PRs, forking, and collaborative work (i.e. what's in the `lec15-git-projects` file). We could link the [Software Carpentry lesson on R projects](https://swcarpentry.github.io/r-novice-gapminder/02-project-intro/index.html) as a guideline for how to structure their repos - I think it's brief and straightforward enough to follow as a guideline. Thoughts?

James-S-Santangelo:
Sounds good. Do you want me to link it in the assignment or do you want to put it in the lecture notes?

PR #279, opened 2018-12-18
File cleanup

aays:
Removes leftover files from the 2017 version of the course, and renames `lec17` to `lec16`.

Resolves UofTCoders/paper-rcourse#19.


lwjohnst86 (approved):
Looks good!

PR #295, opened 2019-05-28
Add Sara to instructor / TA list

mbonsma:


mbonsma:
Fixes https://github.com/UofTCoders/paper-rcourse/issues/72

PR #302, opened 2019-09-03
Updating syllabus based on first instructor meeting

James-S-Santangelo:
- Updated instructors for lessons
- Updated mark breakdown
- Minor updates to course info


about.Rmd
```diff
@@ -17,7 +17,7 @@ Based on [RQM](https://cbahlai.github.io/rqm-template/). Draw much inspiration a
 - Lina Tran (Physiology)
 - James Santangelo (Biology)
```
aays: Since we're listing departments here, should our affiliations be changed to EEB and CSB? Unless this is referring to UTM Biology?

James-S-Santangelo: Done!

PR #303, opened 2019-09-09
Update first two assignments

aays:
Assignment 1:

- Fixed a few typos
- Removed a link to last year's survey and mentioned it'll be on Quercus instead of pasting the new link - prevents us from having to update the link on here every year

Assignment 2:

- Updated the questions to make them slightly more challenging
    - Added a conditional filtering question - it feels a bit silly but I'm not sure how else to test this concept without having them write out a boolean vector of size 30 with `TRUE` over and over - thoughts?
    - Added a debugging question
    - Added a question on `sample_n` - this isn't explicitly discussed in lecture, but the idea is to encourage referencing the cheatsheet for other dplyr functions

### DIFF
--- a/assignment-01.Rmd
+++ b/assignment-01.Rmd
@@ -9,11 +9,11 @@ output:
 knitr::opts_chunk$set(eval = FALSE)
 ```
 
-*To submit this assignment, upload the full document on blackboard,
+*To submit this assignment, upload the full document on Quercus,
 including the original questions, your code, and the output. Submit
-you assignment as a knitted `.pdf` (prefered) or `.html` file.*
+your assignment as a knitted `.pdf` (prefered) or `.html` file.*
 
-1.  Get setup at home (or on a lab computer after hours). (1.5 marks)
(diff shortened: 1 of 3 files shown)


assignment-01.Rmd
```diff
@@ -9,11 +9,11 @@ output:
...
 ```
 
-*To submit this assignment, upload the full document on blackboard,
+*To submit this assignment, upload the full document on Quercus,
 including the original questions, your code, and the output. Submit
-you assignment as a knitted `.pdf` (prefered) or `.html` file.*
+your assignment as a knitted `.pdf` (prefered) or `.html` file.*
 
-1.  Get setup at home (or on a lab computer after hours). (1.5 marks)
+1.  Get set up at home (or on a lab computer after hours). (1.5 marks)
```
James-S-Santangelo, line 16: What about having them print the output of `sessionInfo()` or some equivalent? Basically to confirm that everyone is using the most recent version of R, which they should be.

James-S-Santangelo (commented):
Looks great!

PR #309, opened 2019-09-23
Changes to lectures 4 and 5

aays:
Main change - removing the Team Challenges from lecture 5, since these will instead be used for the new challenge assignment.

I've also pushed some of the material from lecture 4 over to the start of lecture 5, since 4 is fairly long and I haven't been able to cover the full lecture in the two hour slot both years now.


aays:
Uh oh - I know what the issue is. One sec

PR #311, opened 2019-09-24
Update lectures

James-S-Santangelo:
- Fixed typos in my 3 lectures
- Re-ordered (and re-named) lectures and updated lecture cross-references
- Updated references and headers in _site.yml


lec09-Randomization-tests.Rmd
```diff
@@ -265,7 +265,7 @@ ggplot() +
 
 ```
 
-Finally, to get out P-value, we calculate the number of times  the simulated mean difference exceeded the observed mean difference from our data. Because we are performing a two-tailed test, this amounts to determining the number of times the simulated mean difference is **either greater or lesser than the observed difference**. We can do this by asking how many times the absolute value of the simulated mean difference is greater or equal to the absolute value of the observed mean difference.
+Finally, to get out *P*-value, we calculate the number of times  the simulated mean difference exceeded the observed mean difference from our data. Because we are performing a two-tailed test, this amounts to determining the number of times the simulated mean difference is **either greater or lesser than the observed difference**. We can do this by asking how many times the absolute value of the simulated mean difference is greater or equal to the absolute value of the observed mean difference.
```
aays: Not sure about this, but - s/to get out/to get our?

aays: Also double space following 'number of times' near start of line

aays:
As a general comment - could lines be constrained to some character limit (ie 80 chars) so that paragraphs are split over multiple lines? It's a bit hard to see what's been edited in most of these diffs since the entire paragraph gets rendered each time and GitHub doesn't always highlight minor differences in these cases super well.

PR #313, opened 2019-09-25
Lec06

smile4life:
I _think_ I fixed this? Just trying to figure out what I messed up :P


smile4life:
Dangit, meant to make the pull request to a new branch 🤦‍♀

smile4life:
image
I feel like I made a mistake when I rebased?

smile4life:
OKAY. The "files changed" section does only show 4 files. My branch was up-to-date with UofTCoders:master when I made this request 😌 

The rebasing I did for my branch removed the records of those interim commits, I think.

QuLogic:
You rebased too many commits here, e.g., stuff authored by James and committed by you shouldn't be here, since those are redundant.

James-S-Santangelo (changes requested):
Wicked lecture Zoë! I think the students are going to be super excited to finally be handling some real data. Just a couple of quick comments:

1. Ahmed commented on my PR #311 that constraining text to within 80 characters makes it easier for commenting since it avoids having to flag entire paragraphs. The solution is described [here](https://github.com/rstudio/rstudio/issues/2701) and involves reflowing each text paragraph independently by going to Code > Reflow comment in RStudio (Hotkey `Shift + CMD + /` on Mac). The character width is set in Options -> Code -> Display (I used 80).
2. I wonder if it's worth moving the scatterplot section up in the lecture to make sure there is time to cover it. I think it will be helpful for students to get used to seeing scatterplots before the following lecture on linear models. What are your thoughts?

smile4life:
I definitely agree that scatterplots are important to cover, but I'm pretty confident that I'll have time to cover it effectively. I think that most of the live-coding for the earlier sections (mostly just the histogram one?) will just involve making slight adjustments to the first plot: changing the bin widths and filtering the data source?

James-S-Santangelo:
Looks great, thanks Zoë! Will leave it until the morning so that @aays and others have a chance to look through it (Ahmed is currently in a different time zone).

PR #337, opened 2019-11-05
Removing lectures and packages not used in 2019

### LINKED ISSUE #324: Lecture and package removal for 2019
For the next course release, we should remove the following lectures/assignments from the repo since they were replaced for the 2019 course iteration (others may follow):

- assignment-08
- lec12-numerical-models
- lec13-time-series

Similarly, we should remove the following R packages from DESCRIPTION

- phaseR
- mice
- psych
- lattice

James-S-Santangelo:
- Removed `mice`, `phaseR`, and `psych`, which are no longer used in course material. 
- Removed previous lectures on:
    1. Numerical models
    2. Time series
    3. Data viz
    4. Rmarkdown

Should mention this closes: #324


aays:
To be honest, looking back at the reproducibility section of the R Markdown lecture, I kind of would like to see us reintroduce it in some capacity next year to get the students thinking even more about why reproducibility matters but also why working open isn't 'the norm' across all fields quite yet. It may also be worth providing the bits about formatting R Markdown documents as a reference, although I suppose we didn't last year and the students figured things out just fine all the same. Just a thought

PR #339, opened 2019-11-08
Minor edits

James-S-Santangelo:
- Rendering assignment on website


assignment-09-challenge.Rmd
```diff
@@ -0,0 +1,192 @@
...
+---
+
+*To submit this assignment, upload the full document on Quercus,
+including the original questions, your code, and the output. Submit
+your assignment as a knitted `.pdf` (prefered) or `.html` file.*
+
+
+Part of being an effective scientist involves being able to solve problems you
+have not encountered before. This is certainly true of programming as well,
+where problems are typically solved by furious bouts Googling, reading
```
aays: s/bouts Googling/bouts of Googling/

aays: s/bouts Googling/bouts of Googling

assignment-09-challenge.Rmd
```diff
@@ -190,3 +190,56 @@ ggplot(data = simp_df, aes(x = reorder(site, simpson), y = simpson)) +
...
+download.file(plant_biomass_url, 'plant-biomass-preprocess.csv')
+plant_biomass <- read_csv('plant-biomass-preprocess.csv')
+```
+
+Reproduce figure 4 from the paper using `ggplot2`. Pay close attention to the
+overall structure of the figure, the scale on the axes, and so on. The colors of
+the points do not need to be the exact same colors as those in the figure, but
+they should be sufficiently close. _(Note: you can ignore the SEM points, and
+the 'look' of your axes does not have to match the figure exactly. The species
+names also not have to be included in the body of the plot as long as they are
```
aays: s/also not have/also do not have/

aays: Actually - could this final sentence be changed to 'The species names also not have to be included in the body of the plot (like in the paper) as long as they are visible in some form'

assignment-09-challenge.Rmd
```diff
@@ -190,3 +190,56 @@ ggplot(data = simp_df, aes(x = reorder(site, simpson), y = simpson)) +
...
+visible in some form.)_
+
+3. Use the built-in R dataset `iris`, for this question. 
+
+     a. Test the relationship between sepal length and sepal width _for each
+     species_ using linear models. Set sepal width as the response and sepal
+     length as the predictor. Output a single data frame containing the beta
+     estimate and p-value associated with the predictor for each of the three
+     models, and also make sure to include a column called 'species' in the
+     final data frame. Do not include the intercepts. _(Hint: the most efficient
```
aays: Before 'Do not include', can we also add 'Your final data frame should only have species, term, estimate, and p.value as columns.'

assignment-09-challenge.Rmd
```diff
@@ -357,3 +357,111 @@ qplot(x = years, y = lynx) +
...
+vs. drift) affect allele frequencies. In this challenge assignment, we will
+naturally upgrade to a diploid model to see how evolution work in this kind of
+system. 
+
+We will once again be considering the case of malaria (of course, what else does
+Amber think about anyway). In assignment 7, we (you) worked on ways to
+incorporate various facets of biological detail into a baseline, super simple,
+disease model. We will incorporate another level of detail in this exercise:
+host genotype. The trait that we will be working with is the sickle cell trait.
+You can read more about the sickle-cell trait on wikipedia. Basically, this is a
```
aays: s/wikipedia/Wikipedia

assignment-09-challenge.Rmd
```diff
@@ -1,6 +1,8 @@
 ---
 title: "Challenge assignment (14 marks)"
-output: html_notebook
+output:
+  pdf_document: default
+  html_notebook: default
```
aays: So the Travis-breaking error message looks like this:

```
Error in (function (toc = FALSE, toc_depth = 3, toc_float = FALSE, number_sections = FALSE,  : 
  The lib_dir option is not valid for the html_notebook format.
Error: callr subprocess failed: The lib_dir option is not valid for the html_notebook format.
```

I feel like we've dealt with this before, but I forget what the fix was. Did removing `html_notebook` altogether do it?

assignment-09-challenge.Rmd
```diff
@@ -23,85 +25,88 @@ answer to the problems below.
 
 1. Simpson's diversity index and bootstrapping (6 marks)
 
-In lecture 12, we used data from the National Ecological Observatory Networkd on
+    In lecture 12, we used data from the National Ecological Observatory Networkd on
```
aays, line 28: Super minor, but is there a reason for the indent here?

assignment-09-challenge.Rmd
James-S-Santangelo: Yeah I think the yml needs the `html_notebook: default` to be removed

PR #344, opened 2021-01-27
Switch to GitHub Actions for deployment.

### LINKED ISSUE #343: Switch from Travis CI to GitHub Actions
I recently learned that Travis CI is apparently [no longer free](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) and will actually be migrating from travis-ci.org soon. Apparently all projects were also locked and made read-only on Dec 31, 2020. Yikes. 

Apparently [GitHub Actions](https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works) is a viable and free alternative we should consider switching to if future iterations of the course make use of the repo once more. 

A useful reference if we're to do that: https://github.com/geerlingguy/ansible-role-apache/issues/201

QuLogic:
For some reason, it does not build, but there is not much information on the output.

Fixes #343.


QuLogic:
I guess it won't be posting a build here until it's actually merged; here's the build on my repo: https://github.com/QuLogic/rcourse/actions/runs/514062536

It just says it failed to compile, but doesn't really say why.

